PR #26, opened 2020-07-06
Chrome extension

ghost:
## Popup
chrome-extension-1


src/main/webapp/chrome-extension/popup.html
```diff
@@ -0,0 +1,38 @@
...
+        <div class="card" id="gnote-card">
+          <span>Title: <input type="text" id="doc-name-input"></span>
+          <span><input type="checkbox"> Share Publicly</span>
+          <button id="generate-note-btn">Generate</button>
+          <div id="loading">
+            <img src="images/loading.gif">
+          </div>
+        </div>
+      </section>
+      <section id="footer">
```
Aradhya-B, line 30: This isn't a big deal, but since you used the "nav" semantic tag, it's worth noting that there's also a "footer" semantic tag. Would be a good idea to remain consistent and use the semantic tag instead of giving an id="footer".

Aradhya-B (approved):
Looks awesome, great work!

andres-holguin (commented):
Leaving a review so this doesn't show up in my "Awaiting Reviews from You" filter.

PR #29, opened 2020-07-07
Create Vue component for navbar

keelywan:
## Navbar
Screenshot 2020-07-07 at 2 59 57 PM

## Dropdown Menu
Screenshot 2020-07-07 at 3 00 17 PM

- Vue component is in navbar.vue file
  - Custom directive 'click-outside' calls provided method when the user clicks outside the element (used to hide the dropdown menu)
  - Links in the dropdown (e.g. My Profile) will need to be updated when the pages are created
- Navbar.html gives an example of how to load the navbar Vue component by passing in a sample user props


src/main/webapp/components/navbar.vue
```diff
@@ -0,0 +1,158 @@
...
+          <a class="dropdown-link" href="#">My Profile</a> 
+          <a class="dropdown-link" href="#">Favorite Notes</a>
+          <a class="dropdown-link" href="#">Logout</a>
+        </div>
+      </div>
+    </div> 
+  </nav>
+</template>
+
+<script>
```
andres-holguin, line 25: I suggest making component styles scoped so they don't interfere with other similar class names or ids
```suggestion
<script scoped>
```

src/main/webapp/navbar.html
```diff
@@ -0,0 +1,30 @@
...
+          user: {
+            type: Object,
+            default: function() {
+              return {
+                name: 'Bob the Builder',
+                points: 597
+              }
+            }
+          }
+        }
```
andres-holguin: I've never seen props being set for the Vue instance itself before, since there's usually no parent that passes down any information. I might see this be useful if we start using vue's navigation implementation, though. If this prop is only being made for its default value, however, I recommend simply making the object a part of data.
```suggestion
        data: {
          user: {
            name: 'Bob the Builder',
            points: 597
          }
        }
```

src/main/webapp/components/navbar.vue
```diff
@@ -0,0 +1,158 @@
...
+  props: {
+    user: Object
+  },
+  methods: {
+    hideDropdown() {
+      this.showDropdown = false;
+    },
+    toggleDropdown() {
+      this.showDropdown = !this.showDropdown;
+    }
```
andres-holguin, line 55: Love the increase in clarity by making methods for these :)

andres-holguin (changes requested):
This is great, it's super easy to understand for me while also getting a feel for what the common style will be. Once the change to the <style> tag is made, it LGTM 👍

src/main/webapp/components/navbar.vue
```diff
@@ -0,0 +1,158 @@
...
+
+<style>
+.bold {
+  font-weight: bold;
+}
+
+.blue {
+  background-color: #004aad;
+}
+
```
Aradhya-B, line 68: I'm wondering if it might be a good idea to also add a "sticky" class and adding that to the navbar as well to make sure the navbar position remains fixed no matter where the user scrolls.
```suggestion
.sticky { position: fixed; }
```

src/main/webapp/components/navbar.vue
```diff
@@ -0,0 +1,158 @@
...
+  cursor: pointer;
+}
+
+.dropbtn img {
+  height: 65%;
+}
+
+.dropdown-content {
+  align-items: center;
+  background-color: #f9f9f9;
```
Aradhya-B: You have a separate class for the background-color blue above so for consistency it might make sense to make separate classes for this color and in the other selector below. Not too big of a deal though.

src/main/webapp/navbar.html
Aradhya-B: +1

Aradhya-B (approved):
This is awesome Keely, great work!

src/main/webapp/navbar.html
keelywan: I see, I think I misunderstood how to use props. I'll change it to data instead.

src/main/webapp/components/navbar.vue
keelywan: I'll add that!

andres-holguin (approved):

src/main/webapp/components/navbar.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+  <nav class="blue sticky nav-bar">
+    <a href="index.html" class="bold" id="title">starfish</a>
+    <img src="../assets/starfish.png" id="logo">
+    <div id="action-items">
+      <a href="#" id="post-action">PostANote</a>
+      <div class="dropdown">
+        <button class="dropbtn" @click.stop="toggleDropdown">
```
nickKelman: s/dropbtn/dropdown-button

src/main/webapp/components/navbar.vue
```diff
@@ -0,0 +1,167 @@
...
+}
+
+.nav-bar {
+  align-items: center;
+  display: flex;
+  font-family: 'Roboto';
+  height: 80px;
+}
+
+.nav-bar a {
```
nickKelman, line 86: Consider using scss notation to nest css properties:

https://flaviocopes.com/vue-css/

Could be applied to combine this rule with the above rule

nickKelman: Also could apply to some other rules in this file.

nickKelman (approved):
Nice work

src/main/webapp/components/navbar.vue
keelywan: Ok got it, I'll look into adding scss into the project!

andres-holguin: I _think_ vuetify already incorporates scss or sass, but based on [their page on SASS](https://vuetifyjs.com/en/customization/sass-variables/) it's not clear to me how we would use it in our components.

keelywan: Do you mean why we would use scss in our components?

andres-holguin: No, sorry, I meant that since we're not using npm like the page mentions, I can't tell if we need to set up something else before we can start using scss 😆

keelywan: Ohh ok I see, it does seem like we need to do something else to get it set up

src/main/webapp/components/navbar.vue
```diff
@@ -0,0 +1,167 @@
...
+      <a href="#" id="post-action">PostANote</a>
+      <div class="dropdown">
+        <button class="dropdown-btn" @click.stop="toggleDropdown">
+          <img id="user-profile" src="../assets/user.svg">
+        </button>
+        <div class="light-gray dropdown-content" v-if="showDropdown" v-click-outside="hideDropdown">
+          <img src="../assets/user.svg" height="115px"/>
+          <p class="bold" id="dropdown-name">{{ user.name }}</p>
+          <p id="dropdown-points">{{ user.points }} points</p>
+          <a class="dropdown-link" href="#">My Profile</a>
```
cftorres, line 16: I think it's useful to add TODO comments when there are things that will need to be added so you can get into the habit of searching your code for TODOs and ensuring cleanup

cftorres (approved):
Nice work!

PR #31, opened 2020-07-08
This adds a script that can be run to create all the tables in our Cloud SQL MySQL Instance

### LINKED ISSUE #9: Set up database
- [x] Choose storage service
- [x] Create models or tables

### LINKED ISSUE #35: Design DB Schema and make Entity Relationship Diagram

Aradhya-B:
The MySQL instance is up and running and to access it yourselves from the command line, follow the instructions in [this video]
(https://www.youtube.com/watch?v=25XIGXbw_GY).
## Entity Relationship Diagram
image

A note about the size of the VARCHAR datatype in the schema - 255 is the largest number of characters than can be counted with an 8-bit number and is basically the 'max' length for a VARCHAR in most DB systems. In practice, even when set to the max number of characters, VARCHAR still only stores # of bytes + 1 for the text being stored, so it doesn't actually decrease performance. It seems like the only good reason to set a lower character max is for business logic reasons. Check out [this](https://stackoverflow.com/questions/1217466/is-there-a-good-reason-i-see-varchar255-used-so-often-as-opposed-to-another-l) SO post for reference.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.google.sps</groupId>
-  <artifactId>portfolio</artifactId>
+  <artifactId>Starfish</artifactId>
   <version>1</version>
   <packaging>war</packaging>
 
--- a/src/main/sql/sql_tables.sql
+++ b/src/main/sql/sql_tables.sql
@@ -1,0 +1,63 @@
+CREATE TABLE IF NOT EXISTS labels ( 
+  title VARCHAR(255) PRIMARY KEY, 
+  type ENUM('School', 'Course', 'Other') NOT NULL 
+); 
+
+CREATE TABLE IF NOT EXISTS users ( 
+  id INT PRIMARY KEY AUTO_INCREMENT, 
+  display_picture VARCHAR(255), 
+  display_name VARCHAR(255), 
+  date_joined DATE, 
+  email VARCHAR(255) NOT NULL, 
+  points INT NOT NULL, 
+  school VARCHAR(255), 
+  FOREIGN KEY (`school`) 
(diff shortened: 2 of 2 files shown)


andres-holguin (approved):

keelywan (approved):
This looks great!

cftorres (approved):
A thought-- maybe not for the MVP, but... I might consider how to restrict two entries that should be the same when it comes to something like school, for example.  E.g. I think it might be a better UX if you can make it so that UCSD = University of California San Diego

andres-holguin:
> A thought-- maybe not for the MVP, but... I might consider how to restrict two entries that should be the same when it comes to something like school, for example. E.g. I think it might be a better UX if you can make it so that UCSD = University of California San Diego

I definitely agree :) I've started reflecting this in the Front end, but I think we haven't decided how we're going to handle that in the Back end. We started with the idea that School and Course are just like any other label, but have started to deviate from that, it seems.

Aradhya-B:
> A thought-- maybe not for the MVP, but... I might consider how to restrict two entries that should be the same when it comes to something like school, for example. E.g. I think it might be a better UX if you can make it so that UCSD = University of California San Diego

Aradhya-B:
> A thought-- maybe not for the MVP, but... I might consider how to restrict two entries that should be the same when it comes to something like school, for example. E.g. I think it might be a better UX if you can make it so that UCSD = University of California San Diego

I agree as well, some sort of fuzzy matching on the backend to identify labels that are really the same might be able to handle this. E.g. When inserting a new row into the labels table, instead of checking the primary key (which is a VARCHAR(255) title) as a literal string, we could keep an index of fuzzy matches for labels that already exist and check EACH string in that index. If the label to be added already exists in the index, then we wouldn't add it.

PR #33, opened 2020-07-09
This adds a context listener servlet that opens a connection pool to the Cloud SQL Database when an instance of the app spins up

### LINKED ISSUE #9: Set up database

Aradhya-B:
# Description
When the app context is initialized (i.e. AppEngine spins up an instance of our app), the ConnectionPoolContextListener web listener will safely attempt to recreate the entire DB schema on the connected DB instance if the tables don't already exist and add the open connection pool to the app context to be used by all other servlets. This is seen in the ConnectionPoolTestServlet that retrieves the connection pool from the app context and checks if the DB connection is valid.

- DB credentials are stored in environment variables that are configured in the appengine-web.xml config file (that's why I've .gitignored this file for now but in the future we can find a more secure way to save DB credentials)
- Connection pool configurations (maxPoolSize, connectionTimeout, etc.) are pretty standard but can easily be changed if we need to

Check out the active MySQL Instance [here](https://pantheon.corp.google.com/sql/instances/step102-2020-mysql/overview?project=step102-2020).

## Indication of connection to Cloud SQL when running local dev server
image

## Response from test servlet on 'test-pool-connection' path when running local dev server
image


pom.xml
```diff
@@ -33,8 +33,32 @@
...
+      <artifactId>mysql-socket-factory-connector-j-8</artifactId>
+      <version>1.0.16</version>
+    </dependency>
+    <dependency>
+      <groupId>com.zaxxer</groupId>
+      <artifactId>HikariCP</artifactId>
+      <version>3.4.5</version>
+    </dependency>
+    <dependency>
+       <groupId>org.slf4j</groupId>
```
keelywan: nit: I think there might be an additional space before the tags in the last two dependencies.

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,192 @@
...
+    config.setMaxLifetime(1800000); // 30 minutes
+
+    // Initialize the connection pool using the configuration object.
+    DataSource pool = new HikariDataSource(config);
+
+    return pool;
+  }
+
+  private void createLabelsTable(DataSource pool) throws SQLException {
+    try (Connection conn = pool.getConnection()) {
```
keelywan: nit: You have semicolons in some of your try-with-resources but not others. I'm not sure which way is preferred, but just keep it consistent in how it's formatted

pom.xml
```diff
@@ -33,8 +33,32 @@
       <artifactId>appengine-api-1.0-sdk</artifactId>
       <version>1.9.59</version>
     </dependency>
+      <dependency>
```
cftorres: nit: indent out

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,192 @@
...
+  /** Creates every table based on the Starfish DB Schema. */
+  private void createAllTablesInSchema(DataSource pool) throws SQLException {
+    createLabelsTable(pool);
+    createUsersTable(pool);
+    createNotesTable(pool);
+    createOtherNoteLabelsTable(pool);
+    createFavoriteNotesTable(pool);
+  }
+
+  /** This function is called when the Servlet is destroyed. */
```
keelywan: nit: Remove "this function is" so that it's in the same form as the javadoc comment below

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,192 @@
...
+    config.addDataSourceProperty("cloudSqlInstance", CLOUD_SQL_CONNECTION_NAME);
+
+    // maximumPoolSize limits the total number of concurrent connections this pool will keep.
+    config.setMaximumPoolSize(5);
+
+    // minimumIdle is the minimum number of idle connections Hikari maintains in the pool.
+    config.setMinimumIdle(5);
+
+    // setConnectionTimeout is the maximum number of milliseconds to wait for a connection checkout.
+    config.setConnectionTimeout(10000); // 10 seconds
```
cftorres: comments are good but might be even more readable to have a const for a second and min and use multipliers here instead

cftorres (approved):
Awesome work!

keelywan (approved):
This looks great and the comments were super helpful!

pom.xml
Aradhya-B: will do!

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: That's a great point, as is it's not super readable without the comments but with the constants and multipliers it becomes readable either way!

pom.xml
Aradhya-B: You're right, I'll fix that right away.

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: Good point, unfortunately there's nothing on this in the Google Java Style guide but various forums online seem to not use the a semicolon if there's only a single resource being allocated, so to remain consistent I'll remove semicolons in the try-with-resources wherever I've used them.

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: Good catch!

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,195 @@
...
+@WebListener("Creates a connection pool that is stored in the Servlet's context for later use.")
+public class ConnectionPoolContextListener implements ServletContextListener {
+
+  // All database credentials are stored in environment variables
+  private static final String CLOUD_SQL_CONNECTION_NAME = System.getenv("CLOUD_SQL_CONNECTION_NAME");
+  private static final String DB_USER = System.getenv("DB_USER");
+  private static final String DB_PASS = System.getenv("DB_PASS");
+  private static final String DB_NAME = System.getenv("DB_NAME");
+
+  private final int SECOND = 1000;
```
nickKelman: prefer 'MILISECONDS_IN_SECOND'

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,195 @@
...
+  private final int MINUTE = SECOND * 60;
```
nickKelman: prefer 'SECONDS_IN_MINUTE'

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,195 @@
...
+                  + "REFERENCES `labels` (`title`),"
+              + "CONSTRAINT pdf_or_source_url_not_null check ("
+                  + "`source_url` IS NOT NULL OR `pdf_source` IS NOT NULL) );";
+      try (PreparedStatement createTableStatement = conn.prepareStatement(stmt)) {
+        createTableStatement.execute();
+      }
+    }
+  }
+
+  private void createOtherNoteLabelsTable(DataSource pool) throws SQLException {
```
nickKelman: Does 'other' mean non-default here?

Other feels a bit too ambiguous, would prefer a rename to something like non-default.

nickKelman (approved):
Nice work.

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: That is definitely more clear about what this variable actually is, will make that change!

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: I don't think other in this sense exactly means 'non-default'. What it is is that all notes have 2 required labels, which are the school and the course, and then this table represents the bridge between the many to many relationship between the notes and labels for what I called 'other' labels, i.e. labels like 'Physics', 'Chemistry', 'AI', 'Calculus', etc.
Do you think 'optional' would be better, so it would be the 'optional_note_labels' table?

andres-holguin: Or Miscellaneous/Misc, maybe.

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,195 @@
...
+import javax.sql.DataSource;
+
+@WebListener("Creates a connection pool that is stored in the Servlet's context for later use.")
+public class ConnectionPoolContextListener implements ServletContextListener {
+
+  // All database credentials are stored in environment variables
+  private static final String CLOUD_SQL_CONNECTION_NAME = System.getenv("CLOUD_SQL_CONNECTION_NAME");
+  private static final String DB_USER = System.getenv("DB_USER");
+  private static final String DB_PASS = System.getenv("DB_PASS");
+  private static final String DB_NAME = System.getenv("DB_NAME");
```
andres-holguin: So all of these come from the environment, which gets all the env variables by running the script you sent us? How will a deployed server have access to the env? Will we have to deploy it with the same script, somehow?

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: miscLabels definitely sounds better than other and still more descriptive than optional. So I'm thinking I'll push 'createMiscNoteLabelsTable' then?

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,195 @@
...
+
+  private final int MILLISECONDS_IN_SECOND = 1000;
+  private final int SECONDS_IN_MINUTE = SECOND * 60;
+
+  /** Configures database connection, opens pool, and returns pool as DataSource. */
+  private DataSource createConnectionPool() {
+    // The configuration object specifies behaviors for the connection pool.
+    HikariConfig config = new HikariConfig();
```
andres-holguin: I'm not sure what Hikari is. Is it some standard way of setting up JDBC config files?

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: So in the appengine-web.xml file, you can configure the the environment variables on prod instances of your app, so when deployed, the environment will already be configured

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: Basically, but in a more general sense it's straight up a no assembly required JDBC connection pool. You can read more about it on [the official github page](https://github.com/brettwooldridge/HikariCP).

andres-holguin: Interesting, is it something you've used before or that was recommended by the CloudSQL documentation? Don't mean to pry, sorry 😛

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
andres-holguin: Sounds good to me, I can start calling them that in the frontend, too.

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,195 @@
...
+
+    return pool;
+  }
+
+  private void createLabelsTable(DataSource pool) throws SQLException {
+    try (Connection conn = pool.getConnection()) {
+      String stmt =
+          "CREATE TABLE IF NOT EXISTS labels ( "
+              + "title VARCHAR(255) PRIMARY KEY,"
+              + "type ENUM('School', 'Course', 'Other') NOT NULL );";
```
andres-holguin: I'm sure you've thought of this, but I'm hoping there's a way to use sql_tables.sql you merged earlier to simplify the java code. Right now it doesn't seem like this file depends on the .sql at all.

andres-holguin (approved):
LGTM, just a few questions about how things work :)

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: Yep, we should be able to simplify the java code! Was thinking this would be a later thing but we should be able to do what this SO post suggests: https://stackoverflow.com/questions/1044194/running-a-sql-script-using-mysql-with-jdbc

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: I have seen it before but haven't had the opportunity to use it since this is the first time I'm doing CloudSQL stuff in Java. Although it wasn't explicitly mentioned in the CloudSQL documentation, there's a Google Cloud github page where they have examples and I did see them using Hikari to set the config so it seems like it's pretty standard.

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
Aradhya-B: Sounds good!

src/main/java/com/google/starfish/ConnectionPoolContextListener.java
andres-holguin: Very cool, good to know

PR #34, opened 2020-07-10
This adds model classes to represent Notes, Labels, and Users that will be used by the client 

### LINKED ISSUE #9: Set up database

Aradhya-B:
# Description
These classes are a model for the instances that will be sent to client on GET requests for labels, notes, and users. Each instance variable in the class is either a column in the classes respective SQL table, or can be constructed using a structured query (E.g. we can construct an array of miscLabels relating to a specific note from the misc_note_labels table or we can get the count for numFavorites on a note from the favorite_notes table). After this PR is merged, we can start writing clientside code that assumes access to these models on GET requests (just remember to do null checks!) - the servlets and service classes to handle the business logic will be added soon.

**I've included the instance variables I think will be important to the client but may have missed some, so if you think there should be more information on any class please add that to your review!**

*Note 1: Included in this PR is also a table name change in the sql schema file that should have been included with my last open PR but I wasn't able to include because I'd branched off from master before the sql script was merged.*

*Note 2: I'm not 100% sure the best style for long formal parameter lists in class constructors so if you know a better way than what I've done here please let me know!* 

# Update 07/14/20
I've implemented the builder design pattern on the Note and User class and declared non-nullable properties using ```Objects.requireNotNull(obj)```. I also wrote 3 JUnit tests for each class to test the basic constructor functionality and to ensure null pointer exceptions are being thrown when a required parameter is not initialized. See the test files for class usage.

image


src/main/java/com/google/starfish/models/Note.java
```diff
@@ -0,0 +1,55 @@
...
+  private String sourceUrl;
+  private String pdfSource;
+  private Date dateCreated;
+  private long numDownloads;
+  private long numFavorites;
+  private String[] miscLabels;
+
+  /** Default constructor that throws null exception if NOT NULL fields are null */
+  public Note(long id, 
+      long authorId,
```
keelywan: I'm not sure what the best style is either, but I was looking around and it seems that some people start each parameter on a new line (so `long id` would be on it's own line) or you could further indent the other parameters so they start where `long id` starts. Just some ideas 😄 , I don't know if you find either of them more readable.

keelywan (approved):
This looks great!

src/main/java/com/google/starfish/models/Note.java
Aradhya-B: These are good suggestions! I did consider putting the first parameter on a new line as well but that seemed to conflict with what I normally do in this case for html tags so I decided against it. Further indentation to align with the first parameter sounds like a good idea but I'm not 100% sure if it would actually really improve readability in this case. Would also love to hear others' thoughts on this!

src/main/java/com/google/starfish/models/Label.java
```diff
@@ -0,0 +1,21 @@
+package com.google.starfish.models;
+
+/** Enum declaration for type of label. */
+enum Type {
+  SCHOOL,
+  COURSE,
+  OTHER
```
andres-holguin: Recently changed to MISC, I believe

src/main/java/com/google/starfish/models/Label.java
```diff
@@ -0,0 +1,21 @@
...
+
+  private String title;
+  private Type type;
+
+  public Label(String title, Type type) {
+    if (title == null || type == null) { throw new ArgumentNullException(); }
+    this.title = title;
+    this.type = type;
+  }
+}
```
andres-holguin: Note necessary for MVP, but we might want to include a `String code` or `String abbreviation` field for the shorthand names of schools and courses. Maybe SchoolLabels inherit from Label? Also sounds like the DB might need to change a bit. Not sure how to go about this, but definitely not for MVP.

src/main/java/com/google/starfish/models/User.java
```diff
@@ -0,0 +1,45 @@
...
+  private long id;
+  private String displayPicture;
+  private String displayName;
+  private Date dateJoined;
+  private String email;
+  private long points;
+  private String school;
+  // Array of noteIds that have have been favorited by the user
+  private long[] favoriteNotes;
+
```
andres-holguin: I believe we might need an array for uploadedNotes or submittedNotes.

andres-holguin (approved):

andres-holguin:
Looking good 👍

src/main/java/com/google/starfish/models/Label.java
Aradhya-B: This is a good point, we could potentially make that change in the future by making the change to the model and schema but I think you're right that it's not too important for now. I think having SchoolLabels and CourseLabels inheriting from Labels is a good potential solution to add those more specific attributes.

src/main/java/com/google/starfish/models/User.java
Aradhya-B: I think you're right, I'll push that change!

src/main/java/com/google/starfish/models/Label.java
Aradhya-B: Good catch!

src/main/java/com/google/starfish/models/Note.java
nickKelman: Much less familliar with Java but in Javascript we can do the following:
It might be better to pass in a singular variable that is 1 object containing the initializing properties of a note.

This means that when a caller creates a new note, they won't need to insert undefined or null for unused properties and can just define the ones necessary.

Another option could be to use the builder pattern to construct new notes by adding on properties one at a time.

See here for more details on builder pattern:
https://stackoverflow.com/questions/2432443/best-practice-for-passing-many-arguments-to-method

src/main/java/com/google/starfish/models/Note.java
```diff
@@ -0,0 +1,55 @@
...
+      String course, 
+      String title, 
+      String sourceUrl, 
+      String pdfSource, 
+      Date dateCreated, 
+      long numDownloads, 
+      long numFavorites, 
+      String[] miscLabels) {
+
+    if (id == null ||
```
nickKelman: Instead of doing this, could you declare some of the parameters as non-nullable? (Less familiar with Java so genuinely asking)

Aradhya-B: I'm not as familiar with Java either but I just did some research into it and it seems like you can use ```Objects.requireNonNull()```. [Here's the SO Post](https://stackoverflow.com/questions/45632920/why-should-one-use-objects-requirenonnull) I'm referencing. I'll push changes to use this for my parameter Null checking instead of these long if statements because it seems a lot more readable.

src/main/java/com/google/starfish/models/Note.java
Aradhya-B: Very interesting, I think the only way to pass a single object containing the initializing properties of the note in java would be to create a separate data abstraction class for the formal parameter list, which wouldn't really make that big of a difference if ALL the parameters were just clumped into one class. An extension of this is to make multiple data abstraction classes that hunk together related parameters but I feel like creating a bunch of these new classes might just clutter up the client code.
I really like the builder pattern and it seems like you can hunk together related parameters even while using the builder pattern just by passing related parameters in the same builder method. I'll push changes that implement the builder pattern on the note and the user since they have so many instance variables.

cftorres (approved):
Awesome job making use of the Builder pattern and starting to add testing!

src/main/java/com/google/starfish/models/Note.java
nickKelman: Nice looks more readable.

nickKelman (approved):

nickKelman:
Nice job responding to comments and implementing non-trivial refactoring to make the code more readable.

PR #36, opened 2020-07-10
Add some note Components (NoteGrid, NoteGridCollection, and NotePreview)

andres-holguin:
This adds three components that will be used in the generation of notes. NoteGrid and NoteGridCollection serve as containers, while NotePreview is a visual component that shows some Note Data. Styling is subject to change, but the logic of each component is fairly sound. The NoteGridCard component referred to by NoteGridCollection.vue is a WIP, but can be seen behind the preview below.

image

### DIFF
--- a/src/main/webapp/components/NotePreview.vue
+++ b/src/main/webapp/components/NotePreview.vue
@@ -1,0 +1,86 @@
+<template>
+  <v-row justify="center">
+    <v-dialog v-model="showPreview" max-width="860px">
+      <v-card>
+        <v-card-title class="headline">
+          {{title}}
+        </v-card-title>
+        <v-card-subtitle>
+          <em>{{author}}</em> - {{dateString}}
+        </v-card-subtitle>
+
+        <v-badge :content="favoriteCount"
+                  class="ma-2">
+          <v-icon>mdi-star</v-icon>
+        </v-badge>
+        <v-badge :content="downloadCount"
+                  class="ma-2">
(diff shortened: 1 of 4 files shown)


src/main/webapp/components/NotePreview.vue
```diff
@@ -0,0 +1,86 @@
...
+    },
+    mounted: function() {
+      this.$parent.$on('open-preview', note => {
+        this.showPreview = true;
+      });
+    }
+  }
+</script>
+
+<style>
```
keelywan, line 74: nit: I suggest adding scoped to the style tag in case it interferes with other styling

keelywan (approved):
Looks good!

src/main/webapp/components/NotePreview.vue
andres-holguin: Good catch. Unfortunately, the `<v-dialog>` wasn't behaving well to the scoped attribute so I left it out. I will eventually use vuetify components for the .report class so it won't be needed. Not sure what to do about iframe, but I also just tested if this does in fact affect other iframes outside of the component, and it does not. So `scoped` might not even be necessary at all 🤯 .

keelywan: Oh ok that's good to know then!

Aradhya-B (approved):
Great work Andres, and the template should be super useful for future components!

nickKelman (approved):

PR #38, opened 2020-07-13
Create basic upload notes form with Blobstore

keelywan:
## Upload Form
upload-notes

### Servlets
The `UploadNotesServlet` will get the `BlobKey` of the uploaded file. This contains a string identifier which will be stored in the database, and can be used later to retrieve the actual file from Blobstore. (TODO: Connect the servlet with the database and add the form data).

The `ServeNotesServlet` takes the string identifier as a parameter and serves the requested file from Blobstore.

## PDF Preview 
preview-frame

The function `previewUploadedFile` is triggered every time the file input changes, and it will display the user's uploaded file in an iframe.

### DIFF
--- a/src/main/java/com/google/starfish/servlets/UploadNotesServlet.java
+++ b/src/main/java/com/google/starfish/servlets/UploadNotesServlet.java
@@ -1,0 +1,55 @@
+package com.google.servlets;
+
+import com.google.appengine.api.blobstore.BlobInfo;
+import com.google.appengine.api.blobstore.BlobInfoFactory;
+import com.google.appengine.api.blobstore.BlobKey;
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+import com.google.appengine.api.blobstore.BlobInfoFactory;
+import com.google.appengine.api.blobstore.BlobInfo;
+import com.google.appengine.api.images.ServingUrlOptions;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/starfish/servlets/UploadNotesServlet.java
```diff
@@ -0,0 +1,55 @@
...
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String blobKey = getUploadedFileBlobKey(request, "notes-pdf");
+    // TODO: Put blobKey and other course note information into database
+    response.sendRedirect("/");
+  }
+
+  /** Returns a blob key for the uploaded file, or null if the user didn't upload a file. */
+  private String getUploadedFileBlobKey(HttpServletRequest request, String formInputElementName) {
+    BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
```
Aradhya-B: Not sure if you can create a BlobstoreService object outside a method, but if possible, might be good practice to create it as a private field of the Servlet so if there were other methods that used it then they wouldn't have to make an instance of their own BlobstoreService.

Aradhya-B (approved):
Look good Keely, great work!!

src/main/java/com/google/starfish/servlets/UploadNotesServlet.java
keelywan: Yep you can, I'll do that!

andres-holguin (approved):
The logic looks good to me, but I get this error when I submit the form with any of the inputs other than file filled in. 
image
I'm guessing the Servlet just needs to handle the rest of the query string parameters.
Edit: The upload itself works fine; I can visit the /serve-notes and see the PDF. I just get redirected to this error upon submission of the upload form.

nickKelman (approved):
Nice work

PR #39, opened 2020-07-13
Fix image appearance on navbar

keelywan:
This fixes the appearance of the profile image on the navbar when `<!DOCTYPE html>` is added to any html file. 

navbar

Note: In order for the Vue directive `click-outside` to work completely, you will need to add
```css
html,
body {
  height: 100%;
}
```
into your stylesheet. 

Note: This only works when the vuetify stylesheet is also included.

### DIFF
--- a/src/main/webapp/navbar.html
+++ b/src/main/webapp/navbar.html
@@ -4,8 +4,8 @@
     <script src="https://unpkg.com/vue"></script>
     <script src="https://unpkg.com/http-vue-loader"></script>
   </head>
-  <body style="margin: 0 auto">
-    <div id="my-app" style="height: 80px">
+  <body>
+    <div id="my-app">
       <navigation :user="user"></navigation>
     </div>
     <script>
--- a/src/main/webapp/navbar.html
+++ b/src/main/webapp/navbar.html
@@ -1,3 +1,4 @@
+<!DOCTYPE html>
 <html>
   <head>
     <script src="https://unpkg.com/vue"></script>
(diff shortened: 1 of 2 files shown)


Aradhya-B (approved):

src/main/webapp/navbar.html
```diff
@@ -1,11 +1,14 @@
...
     <script src="https://unpkg.com/vue"></script>
     <script src="https://unpkg.com/http-vue-loader"></script>
   </head>
-  <body>
+  <body style="margin: 0 auto">
     <div id="my-app">
-      <navigation :user="user"></navigation>
+      <div style="height:80px">
+        <navigation :user="user"></navigation>
+      </div>
```
andres-holguin: Looks like creating a css reset or normalize stylesheet that we include in all of our pages might be a more sustainable solution. Example: https://meyerweb.com/eric/tools/css/reset/
I'll look into creating one and adding it to template.html

andres-holguin (approved):
Thanks for looking into this :) Btw, I *think* the repo requirement of 3 reviews is only necessary for PRs into master branch. So for small changes like this, it might be better to PR into the bugfixes branch and then we can do one big PR for all bugfixes which gets reviewed by everyone.

andres-holguin:
A few things to add:
* ~I just checked, adding `left: 0` to the `.sticky` class removes the gap around the navbar so `margin: 0` auto isn't needed.~ Unnecessary for `display: sticky`
* I couldn't determine why a `<div>` with height 80px was necessary. Edit: Nevermind, it's to prevent overlap. I'm going to search for a better solution, though. Edit2: try `display: sticky` instead of `display: fixed`.
* ~Since in practice there shouldn't be a page with only the navbar, we probably won't end up needing to manually set  html and body height.~ Vuetify takes care of this

cftorres (approved):

PR #40, opened 2020-07-13
This adds a servlet to handle user registration with Google OAuth and creates a new user session that is saved as a cookie on the user's browser

Aradhya-B:
- New environment variable for the app client id, value can be found [here](https://pantheon.corp.google.com/apis/credentials?project=step102-2020) under Starfish Web Client
- TODO: Test backend auth when connected to client to see if new user is actually created

## Auth Flow
1. Client sends Google Id Token to server after user logs in with Google account
2. Server uses a library to verify the received token
3. If the user doesn't already exist in the database, create a new entry on the user's table
4. Add the user id to a new active session and save the session id as a cookie on the user's browser. All subsequent requests from client will pass this cookie in the header and session will be retrieved by id to see what the corresponding user id is

### DIFF
--- a/src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
+++ b/src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
@@ -1,0 +1,118 @@
+package com.google.starfish.servlets;  
+
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken;
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken.Payload;
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdTokenVerifier;
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import java.io.IOException;  
+import java.io.PrintWriter;  
+import java.sql.Connection;  
+import java.sql.SQLException;  
+import java.sql.PreparedStatement;
+import java.sql.Types;  
+import java.sql.ResultSet;
+import java.sql.Date;
+import javax.servlet.ServletException;  
+import java.security.GeneralSecurityException;
+import javax.servlet.annotation.WebServlet;  
+import javax.servlet.http.HttpServlet;  
+import javax.servlet.http.HttpServletRequest;  
+import javax.servlet.http.HttpServletResponse;  
+import javax.sql.DataSource;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/** Servlet that handles user registration and authentication. */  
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
```diff
@@ -0,0 +1,128 @@
...
+      Payload payload = idToken.getPayload();
+
+      String userId = payload.getSubject();
+
+      // Get profile information from payload
+      String email = payload.getEmail();
+      String name = (String) payload.get("name");
+      String pictureUrl = (String) payload.get("picture");
+      String locale = (String) payload.get("locale");
+      String familyName = (String) payload.get("family_name");
```
keelywan: What is the purpose of `locale` and `familyName`? It doesn't seem like they're being used anywhere else

src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
```diff
@@ -0,0 +1,128 @@
...
+
+          }
+        }
+
+        // Create new user session and save as cookie on client side
+        HttpSession newSession = req.getSession(true);
+        newSession.setAttribute("user_id", userId);
+        Cookie activeSession = new Cookie("sessionId", newSession.getId());
+        res.addCookie(activeSession);
+        res.sendRedirect("/profile.html");
```
keelywan, line 105: Would it be possible to return the user info to the web app (e.g. the number of points a user has or their display name)? Or do you think it'd be better to put that in a separate servlet?

src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
Aradhya-B: Good catch, I just extracted everything I could but if they're not being used anywhere then there's no point of assigning a variable

src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
Aradhya-B: I think we'd put that in a separate servlet, so when you want the user info (maybe onLoad() on profile.html) it would then GET from the /users route and get the user who's id is stored in the active session on the cookie.

keelywan: I think during our discussion today we decided to have the servlet return a User object as a response, in addition to adding the cookie. Do you think you could just add a TODO in the code to say that? And then maybe it doesn't have to redirect to `profile.html` anymore?

andres-holguin (approved):
Nice job, back-end wizard 🧙‍♂️

Aradhya-B:
Closing this PR because there's a new branch that has updated backend auth code that integrates with the frontend.

PR #42, opened 2020-07-14
Add authentication to navbar component

keelywan:
This handles authentication using Google OAuth and changes the appearance of the navbar depending on the authentication status.

## Signed Out Navbar
If the user is not logged in, they can click on the signup/login link.

navbar-signedout

## Popup
A Google sign-in page will popup when the link is clicked.

popup

## Signed In Navbar
Once the user is logged in, they will be able to post notes, view their profile, and logout.

navbar-signedin

### DIFF
--- a/src/main/webapp/components/navbar.vue
+++ b/src/main/webapp/components/navbar.vue
@@ -81,7 +81,6 @@ module.exports = {
     setUserInfo(userData) {
       this.signedIn = true;
       this.user = userData;
-      this.$emit('set-user', userData);
     },
     signIn() {
       this.googleAuth.signIn().then(() => {
@@ -104,6 +103,11 @@ module.exports = {
     gapiScript.onload = this.onGAPILoad;
     gapiScript.src = "https://apis.google.com/js/client.js";
     document.head.appendChild(gapiScript);
+  },
+  watch: {
+    user: function(userVal) {
+      this.$emit('set-user', userVal);
+    }
   }
 }
 </script>
--- a/src/main/webapp/components/navbar.vue
+++ b/src/main/webapp/components/navbar.vue
@@ -84,7 +84,7 @@ module.exports = {
       this.$emit('set-user', userData);
     },
     signIn() {
-      gapi.auth2.getAuthInstance().signIn().then(() => {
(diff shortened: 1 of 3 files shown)


src/main/webapp/components/navbar.vue
```diff
@@ -40,19 +43,67 @@ Vue.directive('click-outside', {
...
+        .then(response => response.json())
+        .then(userData => setUserInfo(userData))
+        .catch(err => {
+          console.log(err);
+        })
+    },
+    setUserInfo(userData) {
+      this.signedIn = true;
+      this.user = userData;
+      this.$emit('set-user', userData);
```
andres-holguin: Thanks for including this line, it should make it easier for the rest of the page to know which user is logged in. Alternatively, you could set a `watch` on `user` so that whenever user changes, that change is `$emit`-ed. This would also ensure that when the user signs out, that change is also emitted, which I don't currently see implemented.

watch example: https://stackoverflow.com/questions/34723680/vuejs-event-on-change-of-element-value

andres-holguin (changes requested):
Nice job :) I'm sure working on this for the chrome extension helped a lot too. I think it's missing a way to announce when the user has signed out, though.

src/main/webapp/components/navbar.vue
keelywan: Nice catch, I completely forgot about the logging out part! And I didn't know that there was a `watch` property, I'll add that instead

andres-holguin:
It might just be me, but I'm getting this error when I try to run the dev server:
image
Changes look good, though.

andres-holguin (approved):

andres-holguin:
Looks like my client needs to get whitelisted? I get this error when I visit navbar.html
image
It cause the component to fail loading correctly, and I can't go through the auth process.
Edit: Whitelisted at [this](https://pantheon.corp.google.com/apis/credentials/oauthclient/506538592562-rueidmib5vvra4cn2ihb48b3fhneb1ka.apps.googleusercontent.com?project=step102-2020) link.

Aradhya-B (approved):
LGTM :), great work!

nickKelman (approved):
Nice work

PR #43, opened 2020-07-15
This adds a servlet to handle basic note search and a NoteService class to handle additional operations of the Note class

### LINKED ISSUE #11: Search for notes (basic)
Implement a rudimentary search for notes based on the school and course. Also allow for sorting based on time of upload.
- [ ] Search by school and/or course
- [ ] Sort by time of upload

Aradhya-B:
# Additions 
- The search servlet reads school and course as query string parameters and returns to the client a list of Note objects as JSON that exactly match the school and course as lowercase strings
- The GenericService class provides basic functionality to get and delete rows in the database by table and id
- The NoteService class inherits from the GenericService class and adds an additional method to get the number of favorites on a note by id, something which is not stored on the notes table itself to ensure consistency with the favorite_notes table

Tests will come in a future PR when testing using a database testing framework has been setup.

# Update 07/17
Changed the type of author id to String in the Note model and test to reflect that change in the DB schema. Also added a new method to NoteService that gets an array of note objects that have been uploaded by a passed user id.


src/main/java/com/google/starfish/servlets/SearchServlet.java
```diff
@@ -0,0 +1,81 @@
...
+  public void doGet(HttpServletRequest req, HttpServletResponse res) throws ServletException, IOException {
+    DataSource pool = (DataSource) req.getServletContext().getAttribute("my-pool");  
+    String reqSchool = req.getParameter("school").toLowerCase();
+    String reqCourse= req.getParameter("course").toLowerCase();
+    ArrayList<Note> notes = new ArrayList<>();
+    try (Connection conn = pool.getConnection()) {
+      String stmt = 
+          "SELECT * "
+        + "FROM `starfish.notes` "
+        + "WHERE `school`= ? AND `course` = ?;";
```
keelywan: I'm not super familiar with SQL but would this statement be able to handle the case that either `school` or `course` might be an empty string?

keelywan (approved):
This looks really good! I like how you created a GenericService then built the NotesService on top of it 😄

src/main/java/com/google/starfish/servlets/SearchServlet.java
Aradhya-B: This is a good question, and I think the answer is no, which might not be exactly what we want. I presume that what should happen if one of school or course is an empty string should be we get all notes with that school or course irrespective of what the empty string field is. Right now, if school or course were empty string, we would probably not get back any notes because right now the query is looking for exact matches. This also makes me think of the possibility of extra whitespace on the search query so I'll push a change that handles that. 
I'm not sure if the empty string problem could be handled without having different querries depending on the case so for now I'll add some conditional logic that changes the statement based on what's empty. If we think of a more efficient way to do this (maybe simply through SQL syntax) we could add that later.

Aradhya-B: Alright I found [this](https://stackoverflow.com/questions/50110968/sql-query-handling-multiple-where-conditions-with-a-potential-null-value) which seems to suggest that the two basic approaches to this problem are dynamic query generation (what I proposed to do with conditional logic) or handle NULL checks in a static SQL statement. Since there aren't many variables I'll use dynamic query generation but in the future if a similar problem arises and the query is more complex, it would make sense to consider handling NULL with a static query.

keelywan: Ok awesome! Also, just letting you know that the search bar component also trims the whitespace. Do you think that should be be removed then if it's redundant?

Aradhya-B: That's good to know! I think we'll leave it in just for safety. I don't think it hurts having those checks on both the frontend and backend.

src/main/java/com/google/starfish/servlets/SearchServlet.java
```diff
@@ -0,0 +1,94 @@
...
+      + "WHERE 1=1";
+    if (school != null && !school.isEmpty()) {
+      stmt += " AND `school`= ?";
+    }
+    if (course != null && !course.isEmpty()) {
+      stmt += " AND `course`= ?";
+    }
+    stmt += ";";
+    return stmt;
+  }
```
andres-holguin, line 78: I'm sure it'll be easy to add a SORT BY dateCreated or numFavorites piece to the statement, but I'm wondering: if SORT BY dateCreated is not specified, do you think the database would return the notes in Chronological order anyway? Is there an order to how rows are stored? Or would omitting SORT BY leave the order unpredictable?

src/main/java/com/google/starfish/servlets/SearchServlet.java
```diff
@@ -0,0 +1,94 @@
...
+          long numDownloads = rs.getLong("num_downloads");
+          long numFavorites = noteService.getNumFavoritesById(pool, noteId);
+
+          Note thisNote = new Note.Builder()
+                              .setId(noteId)
+                              .setAuthorId(authorId)
+                              .setRequiredLabels(school, course)
+                              .setNoteTitle(title)
+                              .setSourceUrl(sourceUrl)
+                              .setOptionalPdfSource(pdfSource)
```
andres-holguin: I know we talked about why we need both of these fields, but right now the frontend only accepts a single pdfSrc field in the Note object since we haven't implemented anything other than pdf storage.

src/main/java/com/google/starfish/services/NoteService.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public class NoteService extends GenericService {
+
+  public ResultSet getById(DataSource pool, long id) throws SQLException {
+    return super.getById(pool, id, Table.NOTES);
+  }
+
+  public boolean deleteById(DataSource pool, long id) throws SQLException {
+    return super.deleteById(pool, id, Table.NOTES);
+  }
```
andres-holguin: Nice use of inheritance 👌

andres-holguin (approved):
Very cool stuff here, yet easy to digest. Excited to see how this works with the search form. Looking forward to those tests.

src/main/java/com/google/starfish/servlets/SearchServlet.java
Aradhya-B: Yeah so we should be able to easily sort by dateCreated or numFavorites but if we don't specify this it's technically undefined, although most likely what will happen is that it returns by time inserted into the database. Like the id field is autoincrementing so technically that's the order in which rows are stored.

src/main/java/com/google/starfish/servlets/SearchServlet.java
Aradhya-B: Sounds good, then I think the Note objects returned to client that just have a pdfSource would hold that in both the sourceUrl and the "optionalPdfSource". This would just be the blobkey, and once we get source URLs for google docs, for any Notes that actually have an associated Google Docs, that's the URL we'd keep in sourceUrl

src/main/java/com/google/starfish/servlets/SearchServlet.java
nickKelman: Great conversation.

nickKelman (approved):
Nice job!

src/main/java/com/google/starfish/services/GenericService.java
```diff
@@ -0,0 +1,57 @@
...
+    return this.sqlTable;
+  }
+
+  private Table(String sqlTable) {
+    this.sqlTable = sqlTable;
+  }
+}
+
+/** Generic Service class that provides the functionality to get and delete row by id in any table */
+public class GenericService {
```
cftorres: nit: maybe naming it something like `TableService` would be more descriptive

src/main/java/com/google/starfish/services/GenericService.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /** Gets row by table and id from sql */
+  public ResultSet getById(DataSource pool, long id, Table sqlTable) throws SQLException {
```
cftorres: nit: I think being a little more verbose like `getRowById` or even just `getRow` (with param name maybe named `rowId` to be a little more explicit) could be better since I might think this is a method for getting a table by ID

src/main/java/com/google/starfish/services/NoteService.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Service class for Notes that allows retrieval of values stored as
+ * instance variables on the Note class that aren't stored on the 
+ * notes table in sql
+ *
+ */
+public class NoteService extends GenericService {
+
+  public ResultSet getById(DataSource pool, long id) throws SQLException {
```
cftorres: Perhaps rather than overriding these methods in each child class, since you're not really adding extra functionality beyond what's in the super method, you could simply require a Table type in the GenericService constructor and save that as an instance variable referenced in the methods rather than taken in as a parameter.  And then here, you can just instantiate super with `Table.NOTES` for example.

src/main/java/com/google/starfish/servlets/SearchServlet.java
```diff
@@ -0,0 +1,94 @@
...
+/** Servlet that returns search results for notes based on school and course */
+@WebServlet("/search")  
+public class SearchServlet extends HttpServlet {  
+
+  private NoteService noteService = new NoteService();
+
+  @Override
+  public void doGet(HttpServletRequest req, HttpServletResponse res) throws ServletException, IOException {
+    DataSource pool = (DataSource) req.getServletContext().getAttribute("my-pool");  
+    String reqSchool = req.getParameter("school").toLowerCase().trim();
```
cftorres: I haven't yet seen where this request is sent, and perhaps you have a safeguard there to make sure "school" and "course" are sent?  But in case you don't, or maybe even if you do (since theoretically this servlet could be reusable in different contexts where, you may forget to require it or so), you might consider extracting this out into a helper method where you verify first that the parameter exists, and then proceed to make it lower case and trim

src/main/java/com/google/starfish/servlets/SearchServlet.java
```diff
@@ -0,0 +1,94 @@
...
+                              .setId(noteId)
+                              .setAuthorId(authorId)
+                              .setRequiredLabels(school, course)
+                              .setNoteTitle(title)
+                              .setSourceUrl(sourceUrl)
+                              .setOptionalPdfSource(pdfSource)
+                              .setDateCreated(dateCreated)
+                              .setNumDownloads(numDownloads)
+                              .setNumFavorites(numFavorites)
+                              .build();
```
cftorres: Perhaps this could be extracted out into a helper method so as not to bloat this servlet method

src/main/java/com/google/starfish/services/GenericService.java
Aradhya-B: I agree, "Generic" sounds very... dare I say, generic.

src/main/java/com/google/starfish/services/GenericService.java
Aradhya-B: Good point, I really like `getRowById`, I think that's still succinct but more descriptive for sure. Will also change param `id` to `rowId`.

src/main/java/com/google/starfish/services/NoteService.java
Aradhya-B: This makes a lot of sense! Since the table isn't changing across methods of the child service, it more clear to keep it as an instance variable in the parent class.

src/main/java/com/google/starfish/servlets/SearchServlet.java
Aradhya-B: This is a really good point, the way it is right now if school or course were null, then we'd get an exception when we try toLowerCase() and trim() it. I think a helper method that does a null check and responds accordingly is the best move.

src/main/java/com/google/starfish/servlets/SearchServlet.java
Aradhya-B: Definitely, we could pass the result set as a param and return an array of note objects!

src/main/java/com/google/starfish/servlets/SearchServlet.java
```diff
@@ -0,0 +1,122 @@
...
+      try {
+        conn.setAutoCommit(false);
+        String stmt = getSearchQuery(reqSchool, reqCourse);
+        try (PreparedStatement getNotesStatement = conn.prepareStatement(stmt)) {
+          getNotesStatement.setString(1, reqSchool);
+          getNotesStatement.setString(2, reqCourse);
+          ResultSet rs = getNotesStatement.executeQuery();
+          conn.commit();
+
+          List<Note> notes = createNotesFromResultOfSeach(pool, rs);
```
cftorres: I think you mean *Search not *Seach (here and in method defn below) :) 

nit: while you're at it though, I think `buildNotesForSearchResults` or `buildNotesFromQueryResult` would be a tad less verbose and a little more accurate than "create" because you're not like creating notes from scratch, you know? And also I think it would be helpful to be a little more consistent in this naming since it sounds like you're using "search" to refer to the user-facing thing, whereas "query" seems to be referring to the internal thing you're doing

cftorres (approved):

src/main/java/com/google/starfish/servlets/SearchServlet.java
Aradhya-B: Ahhh sorry about the typo haha, and I think you're right about renaming, I like `buildNotesFromQueryResult` as this function is taking a query result set as a param.

PR #44, opened 2020-07-15
Create UploadForm.vue

andres-holguin:
## Create Vue Component for filling in and submitting the data that will be stored for each note.

Starfish-Upload-Form-V1

The base UI for the upload form, which can (and probably will) be placed inside a `<v-card>` or `<v-dialog>` component.

This component incorporates the logic made by @keelywan in #38 

### Main features:
* Upon component mount, fetch a new url for uploading a new file
* Upload file (pdf only) and preview it to the side
* Autofill Title field based on file's name.
* Use combobox component to filter through existing items or add a new one for school and course fields
* Prevents a race condition that negatively affects UX
* Send POST request upon form submission to the backend servlet that will store the Note's data.

### Todo:
* Some additional validation logic to be determined later.
* Incorporate into a dialog box that is activated by button on NavBar
* Header for autocomplete menus that communicates that the displayed items are the most common, given conditions.
* A way to pass in or fetch the most common items that appear on the autocomplete menu.

This reinstates previous commit dffc3495448b5eb1caddd56536d765b456ddc6c7 that was reverted for being pushed too early.

### DIFF
--- a/src/main/webapp/components/UploadForm.vue
+++ b/src/main/webapp/components/UploadForm.vue
@@ -1,4 +1,5 @@
 <template>
+  <!-- TODO: Turn into a dialog with a button activator -->
   <v-form ref="form" @submit.prevent="onSubmit">
     <v-row>
       <v-col>
@@ -63,12 +64,14 @@ module.exports = {
       previewUrl: null,
       title: null,
       school: "Unaffiliated",
-      commonSchools: ['UF', 'MIT', 'NYU', 'UCLA'],
+      commonSchools: ['UF', 'MIT', 'NYU', 'UCLA'], // Sample data
       course: "Unaffiliated",
-      commonCourses: ['CIS4301', 'ENC1101', 'PHY2049'],
+      commonCourses: ['CIS4301', 'ENC1101', 'PHY2049'], // Sample data
       miscLabels: null,
-      commonLabels: ['Hard', 'Professor X', 'Test Prep'],
+      commonLabels: ['Hard', 'Professor X', 'Test Prep'], // TODO: Turn into objects so a header can be included.
+                                                          // TODO: fetch commonLabels based on school and course
       noDataHtml: "No matching results. Type and press <kbd>enter</kbd> to create a new one",
+      // TODO: Add rules for form validation
     }
   },
   watch: {
@@ -120,6 +123,7 @@ module.exports = {
     fetch('/blobstore-upload-url')
       .then(response => response.text())
       .then(url => this.uploadUrl = url);
+    // TODO: Fetch the most common schools and courses
   }
 }
 </script>
--- a/src/main/webapp/components/UploadForm.vue
+++ b/src/main/webapp/components/UploadForm.vue
@@ -1,0 +1,131 @@
+<template>
+  <v-form ref="form" @submit.prevent="onSubmit">
+    <v-row>
+      <v-col>
+        <v-file-input label="Upload a PDF"
+                      v-model="file"
+                      accept=".pdf"
+                      show-size
+                      @change="onFileChange">
+        </v-file-input>
+
+        <v-text-field label="Title"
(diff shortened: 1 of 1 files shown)


src/main/webapp/components/UploadForm.vue
```diff
@@ -0,0 +1,131 @@
...
+      this.nextTick(() => this.$refs.school.blur());
+    },
+    course: function() {
+      this.nextTick(() => this.$refs.course.blur());
+    },
+  },
+  methods: {
+    nextTick: function(f) {
+      // Avoids race conditions with function f
+      setTimeout(f, 0);
```
Aradhya-B, line 91: Could you possibly add an image of how the UI is affected by the race condition this is preventing?

Aradhya-B (approved):
Looks great Andres, awesome work! I'd love to see an image of how the UI is negatively affected by the race condition being prevented mostly because I'm curious but if it's not possible to recreate don't worry about it :).

src/main/webapp/components/UploadForm.vue
```diff
@@ -0,0 +1,131 @@
...
+  data: function() {
+    return {
+      uploadUrl: null,
+      file: null,
+      previewUrl: null,
+      title: null,
+      school: "Unaffiliated",
+      commonSchools: ['UF', 'MIT', 'NYU', 'UCLA'],
+      course: "Unaffiliated",
+      commonCourses: ['CIS4301', 'ENC1101', 'PHY2049'],
```
keelywan: Once this is connected to the database, will you need to add a function to retrieve the values for variables like `commonCourses` and `commonLabels`?

keelywan (approved):

src/main/webapp/components/UploadForm.vue
andres-holguin: Yep! It's on the todo. I think there's even a way to show a loading animation while the query results come in, or load them early (eager).

src/main/webapp/components/UploadForm.vue
andres-holguin: As a brief description:
image
Without the timeout, clicking on the MIT suggestion or pressing Enter/Tab at this point wouldn't work. The current search, 'M', would become the label because the input would blur before getting the new MIT suggestion, and the component sets the current search input as the label on blur. So by delaying the blur just a little, the MIT suggestion has time to become the new label.
It's hacky and I don't like it either, but I really needed a way for the suggestions to disappear when a new label was added by the user.

Aradhya-B: Woahhh this is very interesting, I wonder how other sites have implemented it and if they also have to account for this race condition with the blur. I think it's a great solution for the time being since it works and looks epic but is definitely something we could loop back to if we develop a better understanding of how to implement this functionality.

src/main/webapp/components/UploadForm.vue
keelywan: Ok got it, do you think you could add the TODOs into the code as well?

src/main/webapp/components/UploadForm.vue
cftorres: Unfortunately I'm not familiar enough with vue or even native js to know but yes, might be a worthwhile thing to look up, as I imagine you're not the first person doing this!

cftorres (approved):
Nice video in the PR description

src/main/webapp/components/UploadForm.vue
nickKelman: We would do something similar a lot in angularJS using the $timeout service to delay execution of various threads. Thanks for documenting why you are doing this and this could be the recommended way of doing things but's definitely worth double checking.

nickKelman (approved):
Nice work, great PR description. Thanks for giving Keely credit for her work!

PR #45, opened 2020-07-15
This adds servlets to handle user registration and authentication with Google OAuth through cookie and session management

### LINKED ISSUE #12: Register and Log In to account
Implement user registration and authentication with google accounts. Store user data in database.
- [ ] User can register with google account
- [ ] User can sign in and out of account
- [x] User data is stored in the database

Aradhya-B:
# Description of User Registration
Once a user decides to login with their Google Account on the client, the client will send a Google generated OAuth token to the UserRegistrationServlet as a POST request. This servlet verifies the token and if valid, information about the user including their name and email is extracted. If the user doesn't already exist in the database, the user will be inserted using a Google generated ID as the primary key. Once the user is verified and in the database, a new session is created on the server for the user that holds the user's user_id and the sessionId is passed to the client as a cookie. Now, whenever the client wants to manipulate data in the database, servlets will check if the request has an active HttpSession and if that session has a user id (the user id of the client) attached to it. If not, the user isn't logged in yet.
By keeping the userId within the session, the client never needs to pass the userId as part of its request leading to increased security (otherwise, anybody could request data for any user).

# Logout
To logout, the user simply hits the UserLogoutServlet which, if there is an active session, invalidates that session.

# Sessions
Sessions are managed in the HttpSessionCollector which simply keeps track of all active sessions of the app and provides functionality to search for a session by sessionId.

# Testing
A simple test servlet has been written that can be hit by the client to check if they are logged in or not i.e. they have an active session open. More rigorous testing will be implemented once we have a framework in place to test with the database.

# Additional Notes
- To accommodate for the length of Google generated user ids, the type of the user table's primary key (id) has been changed from INT to VARCHAR(255). This is reflected in the sql_tables.sql file and the ConnectionPoolContextListener which creates the tables.
- Once the UserService class is merged (PR coming soon) we will be able to construct user objects and the UserRegistrationServlet will return the user on both a GET and POST request.

### DIFF
--- a/src/main/java/com/google/starfish/servlets/UserLogoutServlet.java
+++ b/src/main/java/com/google/starfish/servlets/UserLogoutServlet.java
@@ -28,6 +28,7 @@ public class UserLogoutServlet extends HttpServlet {
         cookie.setValue(null);
         cookie.setPath("/");
         res.addCookie(cookie);
+        break;
       }
     }
 
--- a/src/main/java/com/google/starfish/servlets/UserLogoutServlet.java
+++ b/src/main/java/com/google/starfish/servlets/UserLogoutServlet.java
@@ -1,0 +1,58 @@
+package com.google.starfish.servlets;  
+
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken;
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken.Payload;
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdTokenVerifier;
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import java.io.IOException;  
+import java.io.PrintWriter;  
+import java.sql.Connection;  
+import java.sql.SQLException;  
+import java.sql.PreparedStatement;
+import java.sql.Types;  
+import java.sql.ResultSet;
+import java.sql.Date;
+import javax.servlet.ServletException;  
+import java.security.GeneralSecurityException;
+import javax.servlet.annotation.WebServlet;  
+import javax.servlet.http.HttpServlet;  
+import javax.servlet.http.HttpSession;  
+import javax.servlet.http.Cookie;  
+import javax.servlet.http.HttpServletRequest;  
+import javax.servlet.http.HttpServletResponse;  
+import javax.sql.DataSource;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/** Servlet that handles user registration and authentication. */  
+@WebServlet("/user-logout")  
+public class UserLogoutServlet extends HttpServlet {
+
+  private static final Logger LOGGER = Logger.getLogger(UserLogoutServlet.class.getName());
+  private final String COOKIE_NAME = "SFCookie";
+
+  @Override 
+  public void doPost(HttpServletRequest req, HttpServletResponse res) throws ServletException, IOException {
+    Cookie[] cookies = req.getCookies();
+    for (Cookie cookie : cookies) {
+      if (COOKIE_NAME.equals(cookie.getName())) {
+        cookie.setMaxAge(0);
+        cookie.setValue(null);
+        cookie.setPath("/");
+        res.addCookie(cookie);
+      }
+    }
+
+    HttpSession activeSession = req.getSession(false);
+    if (activeSession == null) {
+      LOGGER.log(Level.WARNING, "No user is logged in.");
+      // Set an error code of 403 if the user is not logged in
+      res.setStatus(HttpServletResponse.SC_FORBIDDEN);
+      return;
+    }
+    //activeSession.invalidate();
+  }
+}
--- a/src/main/java/com/google/starfish/servlets/UserLogoutServlet.java
+++ b/src/main/java/com/google/starfish/servlets/UserLogoutServlet.java
@@ -1,33 +1,17 @@
 package com.google.starfish.servlets;  
 
-import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken;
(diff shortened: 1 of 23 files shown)


src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
```diff
@@ -0,0 +1,170 @@
...
+          }
+        }
+
+        // Create new user session and save as cookie on client side
+        HttpSession newSession = req.getSession(true);
+        newSession.setAttribute("user_id", userId);
+        Cookie activeSession = new Cookie("SFCookie", newSession.getId());
+        res.addCookie(activeSession);
+
+        // TO-DO: Once we have a UserService, we can create a User object and send that back to client
```
keelywan: Does this mean that after making a POST request to user-signin servlet, it won't be necessary to make a separate GET request to retrieve the User object?

keelywan (approved):
Looks great!

src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
Aradhya-B: Oh good point, I must have forgotten that that's what we'd talked about when I wrote this comment. I think you're still right in that the POST and then GET flow is what makes most sense here, I'll update this comment to reflect that.

src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
```diff
@@ -0,0 +1,170 @@
...
+import javax.servlet.http.HttpServletRequest;  
+import javax.servlet.http.HttpServletResponse;  
+import javax.sql.DataSource;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/** Servlet that handles user registration and authentication. */  
+@WebServlet("/user-signin")
```
andres-holguin: Should this be renamed to match the file name?

andres-holguin (approved):

src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
Aradhya-B: Hm I don't think it's strictly necessary but I can make a change so that it's renamed to user-registration.

src/main/java/com/google/starfish/servlets/UserLogoutServlet.java
```diff
@@ -0,0 +1,44 @@
...
+public class UserLogoutServlet extends HttpServlet {
+
+  private static final Logger LOGGER = Logger.getLogger(UserLogoutServlet.class.getName());
+  private final String COOKIE_NAME = "SFCookie";
+
+  @Override 
+  public void doPost(HttpServletRequest req, HttpServletResponse res) throws ServletException, IOException {
+    // Cookie is unused for now, but will be important once sessions begin to be stored in DB
+    Cookie[] cookies = req.getCookies();
+    for (Cookie cookie : cookies) {
```
nickKelman, line 25: I think this loop will iterate over all cookies. An optimization could be to first find the cookie and then modify it's attributes. This way if the cookie is one of the first, other cookies don't need to be inspected.

src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
```diff
@@ -0,0 +1,170 @@
...
+
+  private static final Logger LOGGER = Logger.getLogger(UserRegistrationServlet.class.getName());
+  private static final String CLIENT_ID = System.getenv("CLIENT_ID");
+  private final String COOKIE_NAME = "SFCookie";
+
+  @Override 
+  public void doGet(HttpServletRequest req, HttpServletResponse res) throws ServletException, IOException {
+    Cookie[] cookies = req.getCookies();
+    String sessionId = null;
+    for (Cookie cookie : cookies) {
```
nickKelman, line 41: Similar to above suggestion, I recommend finding the cookie, caching the value in a local variable, and then using it.

src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
```diff
@@ -0,0 +1,170 @@
...
+            userStmt.setString(3, givenName);
+            userStmt.setDate(4, new Date(Calendar.getInstance().getTimeInMillis()));
+            userStmt.setString(5, email);
+            // User starts with 0 points
+            userStmt.setInt(6, 0);
+            userStmt.setNull(7, Types.VARCHAR);
+
+            // Finally, execute the statement. If it fails, an error will be thrown
+            userStmt.execute();
+
```
nickKelman: nit: extra spacing

nickKelman (approved):
Really nice work. The pull request description was really well written.

src/main/java/com/google/starfish/servlets/UserLogoutServlet.java
Aradhya-B: This is a good point! If we find the Cookie we're looking for we should simply break out of this loop. If the Cookies array was already sorted alphabetically we could have used binary search to go from O(n) to O(logn) but since I don't think it comes to us ordered, it wouldn't be efficient to sort first in O(nlogn) and then run binary search, so I don't think we can do better than O(n) for now. Best thing is probably just to break out of the loop at the end of the IF statement.

src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
Aradhya-B: Sounds good.

src/main/java/com/google/starfish/servlets/UserRegistrationServlet.java
Aradhya-B: Good catch!

PR #46, opened 2020-07-16
Create Blobstore upload servlet and add validation to handle notes servlet 

keelywan:
## BlobstoreUploadServlet
Instead of using a JSP file for the upload notes form to retrieve the Blobstore upload URL, this servlet will return the URL to the form. 

## HandleNotesServlet
In `doPost`, the servlet has validation logic to ensure the user is logged in before they can post a note. It also retrieves the parameters from the upload notes form. 

TODO: Put the notes and notes labels into the database


src/main/java/com/google/starfish/servlets/BlobstoreUploadUrlServlet.java
```diff
@@ -0,0 +1,22 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/blobstore-upload-url")
+public class BlobstoreUploadUrlServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+    String uploadUrl = blobstoreService.createUploadUrl("/upload-notes");
```
andres-holguin:
```suggestion
    String uploadUrl = blobstoreService.createUploadUrl("/handle-notes");
```
Renamed to /handle-notes, right?

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -18,16 +18,35 @@
...
+
+    HttpSession activeSession = request.getSession(false);
+    String userId = (String) activeSession.getAttribute("user_id");
+
+    String blobKey = getUploadedFileBlobKey(request, "file");
+    String title = request.getParameter("title");
+    String school = request.getParameter("school");
+    String course = request.getParameter("course");
+    // TODO: Not sure how the labels will be separated, will need to parse into array of strings
+    String miscLabels = request.getParameter("miscLabels");
```
andres-holguin: Hopefully [getParameterValues()](https://www.java4s.com/java-servlet-tutorials/example-on-getparametervalues-method-of-servlet-request/) will work.
```suggestion
    // TODO: Not sure how the labels will be separated, will need to parse into array of strings
    String[] miscLabels = request.getParameterValues("miscLabels");
```

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -18,16 +18,35 @@
...
+
+    // TODO: Put params into database
+
     response.sendRedirect("/");
```
andres-holguin: I don't think we'll need this redirect if it's not necessary.
```suggestion
```

andres-holguin (approved):

Aradhya-B (approved):
Looks good!

src/main/java/com/google/starfish/servlets/BlobstoreUploadUrlServlet.java
keelywan: Yep, you're right!

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
keelywan: I tried out getParameterValues() and it didn't seem to work, so I think we can just use `split(",")` to turn it into an array.

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -0,0 +1,105 @@
...
+    if(!validateUser(request)) {
+      response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+      return;
+    }
+
+    HttpSession activeSession = request.getSession(false);
+    String userId = (String) activeSession.getAttribute("user_id");
+
+    String blobKey = getUploadedFileBlobKey(request, "file");
+    if(blobKey == null) {
```
nickKelman, line 42: Nice error handling

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * Validates the user using the request's session and cookies. If there is a
+   * valid user logged in, returns true. Otherwise, returns false.
+   **/
+  private boolean validateUser(HttpServletRequest req) {
+    Cookie[] cookies = req.getCookies();
+    String sessionId = null;
+    for (Cookie cookie : cookies) {
```
nickKelman, line 88: Similar to what I put in Aradhya's PR, instead of iterating over all cookies, we should either terminate when we find the correct cookie or use find functionality which I assume there is some library to help with.

nickKelman (approved):
Nice work!

cftorres (approved):
LGTM once cookie logic updated per Nick's comment

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
keelywan: That makes sense, I'll change that!

PR #47, opened 2020-07-17
Add upload form to navbar and make it compatible with Vuetify 

keelywan:
## Navbar
The buttons on the right side of the navbar are now made with Vuetify components. 

vuetify-navbar-signedout

vuetify-navbar-signedin

## Dropdown
The dropdown menu is also made using Vuetify components. 

vuetify-dropdown

## Post A Note
Clicking the Post A Note button will launch the upload form popup.

post-a-note

### DIFF
--- a/src/main/webapp/components/Navbar.vue
+++ b/src/main/webapp/components/Navbar.vue
@@ -1,46 +1,65 @@
 <template>
   <nav class="navy-blue sticky nav-bar">
     <a href="index.html" class="bold" id="title">starfish</a>
-    <img src="../assets/starfish.png" id="logo">
+    <img src="/assets/starfish.png" id="logo">
     <div class="action-items" v-if="!signedIn">
-      <a href="#" class="vert-center" @click.stop="signIn" id="signin-link">Sign Up/Login</a>
+      <v-btn color="primary"
+             dark
+             @click.stop="signIn"
+             id="signin-link">
+        Sign Up/Login
+      </v-btn>
     </div>
     <div class="action-items" v-if="signedIn">
-      <a href="#" class="vert-center">PostANote</a>
-      <div class="dropdown">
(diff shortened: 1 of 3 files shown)


src/main/webapp/components/Navbar.vue
```diff
@@ -1,46 +1,65 @@
...
+             @click.stop="signIn"
+             id="signin-link">
+        Sign Up/Login
+      </v-btn>
     </div>
     <div class="action-items" v-if="signedIn">
-      <a href="#" class="vert-center">PostANote</a>
-      <div class="dropdown">
-        <button class="dropdown-btn" @click.stop="toggleDropdown">
-          <img id="user-profile" src="../assets/user.svg">
```
Aradhya-B, line 12: Should this path be `/assets/user.svg` instead? (Asking because the `..` was removed from the path to `assets` above)

src/main/webapp/components/Navbar.vue
```diff
@@ -1,46 +1,65 @@
...
-        </button>
-        <div class="light-gray dropdown-content" v-if="showDropdown" v-click-outside="hideDropdown">
-          <img src="../assets/user.svg" height="115px"/>
```
Aradhya-B, line 15: Same as above

Aradhya-B (approved):
The vuetify components look epic! Great work!

src/main/webapp/components/Navbar.vue
keelywan: I think all of this code should be deleted, so we actually don't need to worry about it 😄

src/main/webapp/components/Navbar.vue
keelywan: Same thing ^

andres-holguin (approved):
Looks good!

cftorres (approved):

PR #49, opened 2020-07-17
Make search bar compatible with Vuetify

keelywan:
## Search Bar 
The search bar is now made with Vuetify components. 

vuetify-search

## Error
The search bar will also display an error message if both input fields are empty when the search button is clicked.
error-message

### DIFF
--- a/src/main/webapp/components/SearchBar.vue
+++ b/src/main/webapp/components/SearchBar.vue
@@ -1,39 +1,35 @@
 <template>
-  <form class="flex" id="search-container" action="/" @submit.prevent="onSubmit">
-    <div class="flex" id="input-area">
-      <div class="flex" id="search-area">
-        <div class="flex search-item">
-          <label for="school">School:</label>
-          <input
-              type="text"
-              id="school"
-              name="school"
-              v-model.trim="school"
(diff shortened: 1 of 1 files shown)


Aradhya-B (approved):
Man I love Vuetify! Great work!

src/main/webapp/components/SearchBar.vue
```diff
@@ -43,17 +39,19 @@ module.exports = {
       }
       else /* Form validated */ {
         this.error = "";
-        this.queryDatabase(this.school, this.course)
-          .then(result => this.$emit('searchresult', result))
-          .catch(error => this.error = error);
+        this.queryDatabase(this.school, this.course);
```
andres-holguin: Should we handle the error that `queryDatabase()` can throw in some way?

andres-holguin (approved):
Small error handling question, but LGTM otherwise :)

src/main/webapp/components/SearchBar.vue
keelywan: Sure I can add that

cftorres (approved):
All of these Vueify changes look great!

PR #50, opened 2020-07-17
This adds service classes for every DB table that provide methods to servlets to build model instances and retrieve useful information

Aradhya-B:
# Additions
- Add overloaded get and delete by id methods to the parent TableService class so it can be used with the UserService (user ids are Strings, not long)
- Add UserService class that doesn't extend functionality of TableService
- Add LabelService class that extends functionality of TableService by providing public methods to insert a school, course, or misc label
- Add MiscNoteLabelService class that doesn't extend TableService because misc_note_labels uses a compound primary key. Adds standard get, insert, and delete by compound id methods and an additional method that returns an array of labels associated with a given note id
- Add FavoriteNoteService class that doesn't extend TableService because favorite_notes uses a compound primary key. Offers standard get, insert, and delete by compound id methods and two additional methods. One for getting an array of Note objects that have been favorited by a given user, passed by id, and a method to get the number of favorites on any given note, passed by id. The numFavorites method was moved from the NoteService class because it interfaces with the favorite_notes table and not the notes table itself.

### DIFF
--- a/src/main/java/com/google/starfish/services/FavoriteNoteService.java
+++ b/src/main/java/com/google/starfish/services/FavoriteNoteService.java
@@ -67,7 +67,6 @@ public class FavoriteNoteService {
         try (PreparedStatement insertStmt = conn.prepareStatement(stmt)) {
           insertStmt.setLong(1, noteId);
           insertStmt.setString(2, userId);
-          System.out.println(insertStmt);
           insertStmt.execute();
           conn.commit();
         }
--- a/src/main/java/com/google/starfish/services/FavoriteNoteService.java
+++ b/src/main/java/com/google/starfish/services/FavoriteNoteService.java
@@ -7,7 +7,6 @@ import java.sql.ResultSet;
 import javax.sql.DataSource;
 import java.util.List;
 import java.util.ArrayList;
-import java.util.Date;
 import com.google.starfish.models.Note;
 
 /**
@@ -16,6 +15,8 @@ import com.google.starfish.models.Note;
 public class FavoriteNoteService {
 
   private String FAVORITE_NOTES = Table.FAVORITE_NOTES.getSqlTable();
+  private String NOTES = Table.NOTES.getSqlTable();
+  private NoteService noteService = new NoteService();
 
   /** Gets a favorite note by the compoud id */
   public ResultSet getRowByCompoundId(DataSource pool, long noteId, String userId) throws SQLException {
@@ -94,7 +95,7 @@ public class FavoriteNoteService {
       String stmt = 
           "SELECT  a.* "
         + "FROM "
-          + Table.NOTES.getSqlTable() + " AS a "
+          + NOTES + " AS a "
           + "INNER JOIN (SELECT * "
                       + "FROM " + FAVORITE_NOTES + " "
                       + "WHERE user_id=?) "
@@ -103,28 +104,7 @@ public class FavoriteNoteService {
         favNotesStmt.setString(1, userId);
         ResultSet rs = favNotesStmt.executeQuery();
         while (rs.next()) {
-          long noteId = rs.getLong("id");
-          String authorId = rs.getString("author_id");
-          String school = rs.getString("school");
-          String course = rs.getString("course");
-          String title = rs.getString("title");
-          String sourceUrl = rs.getString("title");
-          String pdfSource = rs.getString("pdf_source");
-          Date dateCreated = rs.getDate("date_created");
-          long numDownloads = rs.getLong("num_downloads");
-          long numFavorites = getNumFavoritesByNoteId(pool, noteId);
-
-          Note thisNote = new Note.Builder()
-                              .setId(noteId)
-                              .setAuthorId(authorId)
-                              .setRequiredLabels(school, course)
-                              .setNoteTitle(title)
-                              .setSourceUrl(sourceUrl)
-                              .setOptionalPdfSource(pdfSource)
-                              .setDateCreated(dateCreated)
-                              .setNumDownloads(numDownloads)
-                              .setNumFavorites(numFavorites)
-                              .build();
+          Note thisNote = noteService.constructNoteFromSqlResult(pool, rs);
           notes.add(thisNote);
         }
         rs.close();
--- a/src/main/java/com/google/starfish/services/FavoriteNoteService.java
+++ b/src/main/java/com/google/starfish/services/FavoriteNoteService.java
@@ -7,6 +7,7 @@ import java.sql.ResultSet;
 import javax.sql.DataSource;
 import java.util.List;
 import java.util.ArrayList;
+import java.util.Date;
 import com.google.starfish.models.Note;
 
 /**
@@ -16,7 +17,7 @@ public class FavoriteNoteService {
 
   private String FAVORITE_NOTES = Table.FAVORITE_NOTES.getSqlTable();
   private String NOTES = Table.NOTES.getSqlTable();
-  private NoteService noteService = new NoteService();
+  private MiscNoteLabelService miscNoteLabelService = new MiscNoteLabelService(); 
 
   /** Deletes a favorite note by the compound id */
   public boolean deleteRowByCompoundId(DataSource pool, long noteId, String userId) throws SQLException {
@@ -86,7 +87,7 @@ public class FavoriteNoteService {
         favNotesStmt.setString(1, userId);
         ResultSet rs = favNotesStmt.executeQuery();
         while (rs.next()) {
-          Note thisNote = noteService.constructNoteFromSqlResult(pool, rs);
+          Note thisNote = constructNoteFromSqlResult(pool, rs);
           notes.add(thisNote);
         }
         rs.close();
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -0,0 +1,152 @@
...
+            "INSERT INTO favorite_notes ( "
+                + "note_id,"
+                + "user_id ) "
+          + "VALUES ( "
+                + "?,"
+                + "? ); ";
+        try (PreparedStatement insertStmt = conn.prepareStatement(stmt)) {
+          insertStmt.setLong(1, noteId);
+          insertStmt.setString(2, userId);
+          System.out.println(insertStmt);
```
keelywan: nit: Do you still need the print statement here?

src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -0,0 +1,152 @@
...
+  /** Gets an array of note objects that have been favorited by passed user id */
+  public Note[] getFavoriteNotesByUserId(DataSource pool, String userId) throws SQLException {
+    List<Note> notes = new ArrayList<>();
+    try (Connection conn = pool.getConnection()) {
+      String stmt = 
+          "SELECT  a.* "
+        + "FROM "
+          + Table.NOTES.getSqlTable() + " AS a "
+          + "INNER JOIN (SELECT * "
+                      + "FROM " + Table.FAVORITE_NOTES.getSqlTable() + " "
```
keelywan: nit: I noticed that you used `favorite_notes` in some of your earlier queries then used `Table.FAVORITE_NOTES.getSqlTable()` in this query. Since you already created the Table enum, maybe you could store `Table.FAVORITE_NOTES.getSqlTable()` in a variable and use that? Or just use `favorite_notes` in all the queries?

Same thing with MiscNoteLabelService

src/main/java/com/google/starfish/services/MiscNoteLabelService.java
```diff
@@ -0,0 +1,106 @@
...
+            "INSERT INTO misc_note_labels ( "
+                + "note_id,"
+                + "label ) "
+          + "VALUES ( "
+                + "?,"
+                + "? ); ";
+        try (PreparedStatement insertStmt = conn.prepareStatement(stmt)) {
+          insertStmt.setLong(1, noteId);
+          insertStmt.setString(2, label.toLowerCase().trim());
+          System.out.println(insertStmt);
```
keelywan: nit: same thing as above, do you still need the print statement?

src/main/java/com/google/starfish/services/NoteService.java
```diff
@@ -60,7 +41,7 @@ public long getNumFavoritesById(DataSource pool, long noteId) throws SQLExceptio
           String pdfSource = rs.getString("pdf_source");
           Date dateCreated = rs.getDate("date_created");
           long numDownloads = rs.getLong("num_downloads");
-          long numFavorites = getNumFavoritesById(pool, noteId);
+          long numFavorites = favoriteNoteService.getNumFavoritesByNoteId(pool, noteId);
```
keelywan: Could you add the misc labels for the note here?

src/main/java/com/google/starfish/services/MiscNoteLabelService.java
```diff
@@ -0,0 +1,106 @@
...
+        + "WHERE `note_id`=?;";
+      try (PreparedStatement selectStmt = conn.prepareStatement(stmt)) {
+        selectStmt.setLong(1, noteId);
+        ResultSet rs = selectStmt.executeQuery();
+        List<String> labels = new ArrayList<>();
+        while(rs.next()) {
+          labels.add(rs.getString("label"));
+        }
+        rs.close();
+        return labels.toArray(new String[labels.size()]);
```
keelywan: nit: I thought it would need the size of the ArrayList to construct the array, but I think it might be better to do what you did above and just put 0 instead of `labels.size()`

src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -0,0 +1,152 @@
...
+                              .setAuthorId(authorId)
+                              .setRequiredLabels(school, course)
+                              .setNoteTitle(title)
+                              .setSourceUrl(sourceUrl)
+                              .setOptionalPdfSource(pdfSource)
+                              .setDateCreated(dateCreated)
+                              .setNumDownloads(numDownloads)
+                              .setNumFavorites(numFavorites)
+                              .build();
+          notes.add(thisNote);
```
keelywan, line 91: I noticed that this logic is also used in the NotesService class - do you think it'd be beneficial to make a construct note method or something that would take in the ResultSet and return the note?

src/main/java/com/google/starfish/services/TableService.java
```diff
@@ -97,4 +134,38 @@ public boolean deleteRowById(DataSource pool, long id) {
...
   }
+
+  /** Deletes row by string id from sql */
+  public boolean deleteRowById(DataSource pool, String id) {
+    try (Connection conn = pool.getConnection()) {
+      try {
+        conn.setAutoCommit(false);
+        String stmt = 
+            "DELETE * "
+          + "FROM starfish." + this.sqlTable.getSqlTable() + " "
```
keelywan: nit: I don't think we need the `starfish.` in the query

keelywan (approved):
Looks really good! Can't wait for everything to get onto master 😄

andres-holguin (approved):
LGTM

src/main/java/com/google/starfish/services/FavoriteNoteService.java
Aradhya-B: Nope we don't, I'll remove!

src/main/java/com/google/starfish/services/FavoriteNoteService.java
Aradhya-B: Interesting, yeah I think keeping in a variable is the best approach here since we've already created the enum, I'll make those changes in both files.

src/main/java/com/google/starfish/services/MiscNoteLabelService.java
Aradhya-B: Nope!

src/main/java/com/google/starfish/services/NoteService.java
Aradhya-B: Yes will do that.

src/main/java/com/google/starfish/services/MiscNoteLabelService.java
Aradhya-B: Ahhh I see, yeah I must've missed this when I was adding your function. Good catch!

src/main/java/com/google/starfish/services/FavoriteNoteService.java
Aradhya-B: Yes potentially, I think it could be a method in the NoteService class and then it's used as a dependency in the FavoriteNoteService class. I'll make a change.

src/main/java/com/google/starfish/services/TableService.java
Aradhya-B: You're right, good catch!

cftorres (approved):
Discussed offline-- based on my understanding of your product's user journeys, it doesn't seem like getting a specific favorite note or a specific label is ever really needed, so you might consider leaving these methods out unless you're sure you're going to use them (following the YAGNI principle) because it's easy enough to add functionality after anyway.  But, really awesome DB work!

cftorres (approved):

src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -0,0 +1,144 @@
...
+        return notes.toArray(new Note[0]);
+      }
+    }
+  }
+
+  /** Gets the number of times a note has been favorited by note id */
+  public long getNumFavoritesByNoteId(DataSource pool, long noteId) throws SQLException {
+    try (Connection conn = pool.getConnection()) {
+      String stmt = 
+          "SELECT COUNT(*) AS num_favorites "
```
nickKelman: Given that 'num_favorites' is an id you use later in this function, I'd cache it in a variable and use it twice so they never go out of sync.

src/main/java/com/google/starfish/services/MiscNoteLabelService.java
```diff
@@ -0,0 +1,89 @@
...
+    } catch (SQLException ex) {
+      System.err.print(ex);
+    }
+  }
+
+   /** Gets the misc labels attached to the note with the given note id */
+  public String[] getMiscLabelsByNoteId(DataSource pool, long noteId) throws SQLException {
+    try(Connection conn = pool.getConnection()) {
+      String stmt =
+          "SELECT label "
```
nickKelman: Above comment applies here.

src/main/java/com/google/starfish/services/NoteService.java
```diff
@@ -11,73 +11,63 @@
...
-                              .setAuthorId(authorId)
-                              .setRequiredLabels(school, course)
-                              .setNoteTitle(title)
-                              .setSourceUrl(sourceUrl)
-                              .setOptionalPdfSource(pdfSource)
-                              .setDateCreated(dateCreated)
-                              .setNumDownloads(numDownloads)
-                              .setNumFavorites(numFavorites)
-                              .build();
+          Note thisNote = constructNoteFromSqlResult(pool, rs);
```
nickKelman: Can inline this variable since used once and should be a short line.

src/main/java/com/google/starfish/services/NoteService.java
```diff
@@ -11,73 +11,63 @@
...
           notes.add(thisNote);
         }
         rs.close();
         return notes.toArray(new Note[0]);
```
nickKelman, line 40: Out of curiosity, why do we need to pass in a new Note object to this toArray function?

nickKelman (approved):
Nice work!

src/main/java/com/google/starfish/services/FavoriteNoteService.java
Aradhya-B: Good catch!

src/main/java/com/google/starfish/services/NoteService.java
Aradhya-B: Since this is being done inline (i.e. I'm not caching what I'm returning in a local variable first), you have to pass in the type of the array the array list is being converted into, so in this case I'm basically saying "convert this arraylist to a normal array that holds `Note` objects".
The alternative here is cache as a local variable first (which will declare the type of the array) and then return, in which case I could also do:
```
Note[] notesArray = notes.toArray();
return notesArray;
```

PR #53, opened 2020-07-20
Change chrome extension authentication and add generate note functionality

### LINKED ISSUE #24: Generate a GNote
Users can generate GNotes from the site. Users can pick from a handful of templates that should improve note-taking quality. Users can also enter metadata so it gets generated with the GNote.

keelywan:
## Authentication 
`gapi.auth2` is used for authentication on the webapp, but it is not supported by chrome extensions. Because of this, we have modified the authentication of the chrome extension by using `chrome.identity.launchWebAuthFlow`. (Previously, we used `chrome.identity.getAuthToken` which gave a different type of access token). Part of the response from `chrome.identity.launchWebAuthFlow` will include an authorization code, which is then included in the url to retrieve the access token through `getTokenEndpoint`. This gives us the same access token that is generated through `gapi.auth2` in the webapp. 

## Signed Out Popup
When the user is signed out, there will be a login button in the footer and the generate note button will be disabled.
signed-out-popup

## Signed In Popup
When the user is signed in, the logout button will appear in the footer, the user email will be displayed, and the generate note button will be enabled. 
signed-in-popup

## Loading Icon
The loading icon will appear after the user clicks Generate, and it will disappear after the new Google Doc is created. The chrome extension will then open the Google Doc in a new tab.
loading

Note: `idToken` is the token used by the backend servlets to verify the user, `accessToken` is the token used by gapi to call the APIs like Google Drive and Google Docs

TODO 1: Make a POST request to the user servlet when logging in
TODO 2: Make a POST request to the upload servlet when generating a new gNote


Aradhya-B (approved):
This looks really good! This isn't too big of a deal, but generally speaking, there seems to be a lot of inconsistency in when you're using regular functions and arrow functions. There's even a case where in a single promise chain you've used a regular function in one `then()` and then an arrow function in the next `then()`. I don't think one type is necessarily better than the other, especially in the context of promise chains, but it would be a good idea to use one or the other in the same context for the sake of consistency.

andres-holguin (approved):

chrome-extension/popup.js
```diff
@@ -23,79 +42,139 @@ window.onload = function() {
...
+      '&response_type=code' + 
+      '&access_type=offline' +
+      '&redirect_uri=' + REDIRECT_URI + 
+      '&scope=' + SCOPES;
+  chrome.identity.launchWebAuthFlow({
+    url: url,
+    interactive: true
+  }, redirectedTo => {
+    if(chrome.runtime.lastError) {
+      console.log(chrome.runtime.lastError);
```
nickKelman: Typically its good to remove log statements used for debugging into the master branch.

nickKelman (approved):

PR #55, opened 2020-07-20
This adds functionality to HandleNotesServlet to actually upload the posted note and its associated labels to the database

### LINKED ISSUE #10: Upload notes
Store in the database any PDF of notes from the form the user fills out on the upload-note page.
- [x] Handle PDF uploads and GDoc links differently based on existence of Blobkey
"If the blobkey is null, then it's a Gdoc. For the check, see if both blobkey and sourceUrl are null. Then return status error if they are both null" -Keely.

Aradhya-B:
# Ingest Flow
- Validate that the user is logged in and get their `userId` (must be logged in to upload a note)
- Check if there is a `sourceUrl` passed as a param or a valid `blobKey` for an uploaded file: if there isn't, do nothing and return
- Get all remaining params that describe the note
- Get all labels (`school`, `course`, `misc`)
- Insert all labels into the LABELS table because of FK restraint on the NOTES table
- Insert the new Note into the NOTES table and get the newly posted `noteId`
- Associate the new Note with each misc label by inserting a new row into the MISC_NOTE_LABELS table for every misc label

## Note About null checks on `blobKey` and `sourceUrl`
On the NOTES table, `sourceUrl` is not null and `blobKey` is nullable. The original design decision made was to make it so either of these columns could be `null`, but both couldn't be `null` at the same time. Unfortunately, this is [not possible to do in MySQL](https://stackoverflow.com/questions/33611597/how-define-column-a-or-column-b-not-null) (Although it is possible in SQLServer). 
Realizing this, the alternative usage considers the two note uploading cases: "Uploading" a Google Doc, and uploading a PDF.

### Uploading a Google Doc
When uploading a Google Doc, the `sourceUrl` must be passed as a query param and must be not null. In this case however, there will be no blob key (because there is no associated PDF), so in the DB, the `pdfSource` will be `null` (which is fine because pdfSource is nullable). This nicely sets up the possibility for future functionality when a Google Doc could be transformed into a PDF on the platform and we'd want to keep the Google Doc url in `sourceUrl` and the `blobKey` in `pdfSource`.

### Uploading a PDF
When uploading a PDF, the servlet will get a `blobKey `itself by handling the uploading to Blobstore. The only "source" for this PDF is Blobstore, so it makes sense to keep the `blobKey` as the `sourceUrl` in addition to keeping the `pdfSource` the `blobKey` as well.

### DIFF
--- a/src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
+++ b/src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
@@ -25,7 +25,7 @@ import java.sql.ResultSet;
 import java.sql.Date;
 import java.util.Calendar;
 
-
+/** Servlet to handle note uploading and adding associated labels to the database */
 @WebServlet("/handle-notes")
 public class HandleNotesServlet extends HttpServlet {
 
--- a/src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
+++ b/src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
@@ -5,6 +5,8 @@ import com.google.appengine.api.blobstore.BlobInfoFactory;
 import com.google.appengine.api.blobstore.BlobKey;
 import com.google.appengine.api.blobstore.BlobstoreService;
 import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+import com.google.protos.cloud.sql.Client.SqlException;
+
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
@@ -60,12 +62,7 @@ public class HandleNotesServlet extends HttpServlet {
     DataSource pool = (DataSource) request.getServletContext().getAttribute("my-pool");
 
     // Insert all labels before inserting note due to foreign key constraint
-    labelService.insertSchoolLabel(pool, school);
-    labelService.insertCourseLabel(pool, course);
-    for (String miscLabel : miscLabels) {
-      labelService.insertMiscLabel(pool, miscLabel.toLowerCase());
-    }
-    
+    addLabels(pool, school, course, miscLabels);
     try (Connection conn = pool.getConnection()) {
       String stmt =
           "INSERT INTO notes ( "
@@ -92,30 +89,13 @@ public class HandleNotesServlet extends HttpServlet {
         noteStmt.setString(2, school);
         noteStmt.setString(3, course);
         noteStmt.setString(4, title);
-        if (sourceUrl != null) {
-          // If there is a sourceUrl on the request param, this is
-          // a google doc set the source url based on the passed param
-          noteStmt.setString(5, sourceUrl);
-        } else {
-          // If there was no sourceUrl, this is an uploaded pdf
-          // so set the source url to serve-notes path
-          noteStmt.setString(5, "/serve-notes?key=" + blobKey);
-        }
-        // The blob key is nullable on the notes table so if this is a google doc
-        // this will be null
+        noteStmt.setString(5, validateAndGetSourceUrl(sourceUrl, blobKey));
         noteStmt.setString(6, blobKey);
         noteStmt.setDate(7, new Date(Calendar.getInstance().getTimeInMillis()));
         int rowsAffected = noteStmt.executeUpdate();
         if (rowsAffected == 1) {
-          long noteId = 0;
-          ResultSet rs = noteStmt.getGeneratedKeys();
-          if (rs.next()) {
-            noteId = rs.getLong(1);
-            // Associate misc labels to the newly added note
-            for (String miscLabel : miscLabels) {
-              miscNoteLabelService.insertMiscNoteLabel(pool, noteId, miscLabel);
-            }
-          } 
+          ResultSet generatedKeys = noteStmt.getGeneratedKeys();
+          addMiscLabelsToNewNote(pool, generatedKeys, miscLabels);
         }
       }
     } catch (SQLException ex) {
@@ -123,6 +103,39 @@ public class HandleNotesServlet extends HttpServlet {
       response.getWriter().println("INTERNAL SERVER ERROR: " + ex);
     }    
   }
+
+  /** Inserts a school, course, and array of misc labels to the database */
+  private void addLabels(DataSource pool, String school, String course, String[] miscLabels) {
+    labelService.insertSchoolLabel(pool, school);
+    labelService.insertCourseLabel(pool, course);
+    for (String miscLabel : miscLabels) {
+      labelService.insertMiscLabel(pool, miscLabel.toLowerCase());
+    }
+  }
+
+  /** Associates an array of misc labels with a newly added Note in the database */
+  private void addMiscLabelsToNewNote(DataSource pool, ResultSet generatedKeys, String[] miscLabels) throws SQLException {
+    if (generatedKeys.next()) {
+      long noteId = generatedKeys.getLong(1);
+      for (String miscLabel : miscLabels) {
+        miscNoteLabelService.insertMiscNoteLabel(pool, noteId, miscLabel);
+      }
+    }
+  }
+
+  /** Returns the appropriate source_url string for a note after determining whether 
+   *  the note is a Google Doc or a PDF
+   */
+  private String validateAndGetSourceUrl(String sourceUrl, String blobKey) {
+    if (sourceUrl != null) {
+      // If there is a sourceUrl on the request param, this is
+      // a google doc set the source url based on the passed param
+      return sourceUrl;
+    }
+    // If there was no sourceUrl, this is an uploaded pdf
+    // so set the source url to serve-notes path
+    return "/serve-notes?key=" + blobKey;
+    }
 
   /** Returns a blob key for the uploaded file, or null if the user didn't upload a file. */
   private String getUploadedFileBlobKey(HttpServletRequest request, String formInputElementName) {
--- a/src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
+++ b/src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
@@ -123,7 +123,6 @@ public class HandleNotesServlet extends HttpServlet {
       response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
       response.getWriter().println("INTERNAL SERVER ERROR: " + ex);
     }    
-    response.sendRedirect("/");
   }
 
   /** Returns a blob key for the uploaded file, or null if the user didn't upload a file. */
--- a/src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
+++ b/src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
@@ -55,7 +55,7 @@ public class HandleNotesServlet extends HttpServlet {
     String title = request.getParameter("title");
     String school = request.getParameter("school").toLowerCase();
     String course = request.getParameter("course").toLowerCase();
-    String[] miscLabels = request.getParameter("miscLabels").split(",");
+    String[] miscLabels = request.getParameterValues("miscLabels");
 
     DataSource pool = (DataSource) request.getServletContext().getAttribute("my-pool");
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -1,58 +1,129 @@
...
+              miscNoteLabelService.insertMiscNoteLabel(pool, noteId, miscLabel);
+            }
+          } 
+        }
+      }
+    } catch (SQLException ex) {
+      response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+      response.getWriter().println("INTERNAL SERVER ERROR: " + ex);
+    }    
+    response.sendRedirect("/");
```
keelywan: I don't think we need to the redirect anymore since the form will just close after it is submitted

keelywan (approved):
Looks great!

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
Aradhya-B: Ahh good point, I'll remove it then!

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -1,58 +1,128 @@
...
+    if (blobKey == null && sourceUrl == null) {
       response.setStatus(HttpServletResponse.SC_NOT_ACCEPTABLE);
       return;
     }
-    String title = request.getParameter("title").toLowerCase();
+
+    String title = request.getParameter("title");
     String school = request.getParameter("school").toLowerCase();
     String course = request.getParameter("course").toLowerCase();
     String[] miscLabels = request.getParameter("miscLabels").split(",");
```
andres-holguin: I should probably make sure the front end doesn't allow commas in label names, then. But how about [getParameterValues](https://www.java4s.com/java-servlet-tutorials/example-on-getparametervalues-method-of-servlet-request/), instead? This is a bug waiting to happen, ATM.

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -1,58 +1,128 @@
...
+        } else {
+          // If there is a sourceUrl on the request param, this is
+          // a google doc set set the source url based on the passed param
+          noteStmt.setString(5, sourceUrl);
+        }
+        // The blob key is nullable on the notes table so if this is a google doc
+        // this will be null
+        noteStmt.setString(6, blobKey);
+        noteStmt.setDate(7, new Date(Calendar.getInstance().getTimeInMillis()));
+        noteStmt.setLong(8, 0);
```
andres-holguin:
```suggestion
        noteStmt.setLong(8, 0); // num_downloads
```

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -88,7 +158,6 @@ private boolean validateUser(HttpServletRequest req) {
     for (Cookie cookie : cookies) {
       if (COOKIE_NAME.equals(cookie.getName())) {
         sessionId = cookie.getValue();
-        break;
```
andres-holguin: Isn't this an addition from another PR review? There might be some changes that need to be merged.

andres-holguin (changes requested):
Found a potential bug, with a simple fix (I hope). Very nice, though.

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
keelywan: I tried testing getParameterValues once, and it seemed to just combine all the labels together. I'm not sure if I was doing it wrong, but maybe we could try it again.

Aradhya-B: From what I tested, `.split(",")` seemed to be working ok but I see how this could easily become a bug. `getParameterValues` definitely looks like the right solution here so I'll make that change, good catch!

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -1,58 +1,128 @@
...
+
+      try (PreparedStatement noteStmt = conn.prepareStatement(stmt, new String[] {"id"})) {
+        noteStmt.setString(1, userId);
+        noteStmt.setString(2, school);
+        noteStmt.setString(3, course);
+        noteStmt.setString(4, title);
+        if (sourceUrl == null) {
+          // If there was no sourceUrl, this is an uploaded pdf so
+          // set the sourceUrl to be the same as the pdfSource
+          noteStmt.setString(5, blobKey);
```
andres-holguin: Currently, the front end coverts the blobKey to a URL with simply:
` src = "/serve-notes?key=" + srcUrl`
I think it would make more sense if the back-end did this conversion of blobKey to url, so that:
1. This line's only named variable matches the table field it is setting (`blobKey` -> `srcUrl`)
2. The front end can simply use srcUrl instead of having to convert it to src
```suggestion
         // Convert blobKey to srcUrl
          noteStmt.setString(5, srcUrl);
```
The notion of blobKeys should be completely obscured in the front end.

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
Aradhya-B: Ohh you are right, I think just adding the break back to this PR will prevent it from breaking anything on merge and keep what we want. Good catch!

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
Aradhya-B: This makes a lot of sense! Even if it's an uploaded PDF, we should be keeping an actual source url as the `sourceUrl` in our DB, not just the blobKey as then the usage of `sourceUrl` would become convoluted otherwise. I'll make the change!

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
andres-holguin: @keelywan sounds like maybe you did

`String values = getParameterValues(“Input Parameter”);`

instead of

`String[] values = getParameterValues(“Input Parameter”);` 
?

andres-holguin (approved):
Thanks for making those changes :)

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -1,58 +1,128 @@
...
+              + "?,"
+              + "?,"
+              + "? ); ";
+
+      try (PreparedStatement noteStmt = conn.prepareStatement(stmt, new String[] {"id"})) {
+        noteStmt.setString(1, userId);
+        noteStmt.setString(2, school);
+        noteStmt.setString(3, course);
+        noteStmt.setString(4, title);
+        if (sourceUrl == null) {
```
nickKelman: It feels odd to check for sourceUrl as the requirement to add a blobkey. Could we instead check for a sourceUrl to set the source_url as sourceUrl?

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -1,58 +1,128 @@
...
+          // If there was no sourceUrl, this is an uploaded pdf
+          // so set the source url to serve-notes path
+          noteStmt.setString(5, "/serve-notes?key=" + blobKey);
+        } else {
+          // If there is a sourceUrl on the request param, this is
+          // a google doc set set the source url based on the passed param
```
nickKelman: nit: s/set set/set

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -1,58 +1,128 @@
...
+          noteStmt.setString(5, sourceUrl);
+        }
+        // The blob key is nullable on the notes table so if this is a google doc
+        // this will be null
+        noteStmt.setString(6, blobKey);
+        noteStmt.setDate(7, new Date(Calendar.getInstance().getTimeInMillis()));
+        noteStmt.setLong(8, 0); // num_downloads
```
nickKelman: if numdownloads is always 0, we don't need to dynamically set it here, we can just set it in the prepared statement.

nickKelman (approved):

nickKelman (approved):

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
Aradhya-B: Yep, I'll check for `sourceUrl != null` instead!

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
Aradhya-B: Good point, I'll make that change

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -1,58 +1,128 @@
...
+        noteStmt.setString(3, course);
+        noteStmt.setString(4, title);
+        if (sourceUrl == null) {
+          // If there was no sourceUrl, this is an uploaded pdf
+          // so set the source url to serve-notes path
+          noteStmt.setString(5, "/serve-notes?key=" + blobKey);
+        } else {
+          // If there is a sourceUrl on the request param, this is
+          // a google doc set set the source url based on the passed param
+          noteStmt.setString(5, sourceUrl);
```
cftorres: since you have this whole set of things you're setting and it's already a little lengthy, I might consider extracting this logic out as well into a helper so you can make a singular call here amidst all the others

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -1,58 +1,128 @@
...
+        if (rowsAffected == 1) {
+          long noteId = 0;
+          ResultSet rs = noteStmt.getGeneratedKeys();
+          if (rs.next()) {
+            noteId = rs.getLong(1);
+            // Associate misc labels to the newly added note
+            for (String miscLabel : miscLabels) {
+              miscNoteLabelService.insertMiscNoteLabel(pool, noteId, miscLabel);
+            }
+          }
```
cftorres: This looks like something too you can extract out into a helper so that you can just make a simple call here like `addLabelsToNote()`

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -1,58 +1,128 @@
...
+    DataSource pool = (DataSource) request.getServletContext().getAttribute("my-pool");
 
-    response.setStatus(HttpServletResponse.SC_OK);
+    // Insert all labels before inserting note due to foreign key constraint
+    labelService.insertSchoolLabel(pool, school);
+    labelService.insertCourseLabel(pool, course);
+    for (String miscLabel : miscLabels) {
+      labelService.insertMiscLabel(pool, miscLabel.toLowerCase());
+    }
+
```
cftorres: Even though it doesn't seem like much, I might extract this out into a helper to keep this servlet method as condensed as possible since there's already quite a lot going on below that can't really be condensed

cftorres (approved):

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
Aradhya-B: Good point, that would make it a lot cleaner! I'll make a helper function called `validateAndGetSourceUrl` or something which handles the logic to set the correct source_url.

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
Aradhya-B: Yep definitely! That would get rid of this extra bloating.

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
Aradhya-B: You're definitely right here, there's a lot going on and the more that can be abstracted out the more readable and maintainable this code becomes!

PR #54, opened 2020-07-20
Fix minor bugs

keelywan:
This fixes the bugs in SearchServlet, NoteGridCard, and SearchBar so that the MVP demo will work.

PR #56, opened 2020-07-20
This adds functionality to the UserRegistrationServlet to return a User object for the logged in user on GET

### LINKED ISSUE #13: View Profile page
Load up the profile based on the user data in the database. Profile for _username_ can be found on /_username_.

- [x] Return User object on GET request from user servlet (Aradhya)
- [x] Take current mock and modify assuming you have a user object.

Aradhya-B:
# Changes
- The `id` field in the User class is now String instead of long to reflect earlier schema changes
- `uploadedNotes` and `favoriteNotes` are now `Note` type arrays instead of `long` type arrays (previously they would just hold note ids but now they hold the entire Note object)
- onGet in `user-registration` will return the object representation of the currently logged in user which will include all information needed to construct the user's personal profile page (notably the user's favorited and uploaded notes are completely included in the response json)
- Constructor tests for the User class have been modified to reflect the changes made to the User class

# Example JSON Response onGet when Aradhya is logged in
image

*Note: I merged #54 to this branch when I first branched to do testing but removed those changes because #54 is still outstanding at the time I'm opening this PR.*

### DIFF
--- a/src/main/java/com/google/starfish/models/User.java
+++ b/src/main/java/com/google/starfish/models/User.java
@@ -6,7 +6,7 @@ import java.util.Objects;
 /** Represents a user */
 public class User {
 
-  private long id;
+  private String id;
   private String displayPicture;
   private String displayName;
   private Date dateJoined;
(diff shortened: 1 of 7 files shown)


keelywan (approved):
LGTM!

andres-holguin (approved):
Cleeeeeean 🧼

nickKelman (approved):

cftorres (approved):

PR #57, opened 2020-07-20
This adds a servlet that can return all school and course labels as json

### LINKED ISSUE #17: Fetch existing schools and courses
When a user is adding a new label, a drop-down should pop up with suggestions for already-existing labels based on what the user has typed so far. 
Example:
**Labels in Existence**: ["Red", "Blue", "Yellow", "Purple", "Pink"]
**When the user has typed**:

- **Nothing**, no more than about 3 suggestions should drop down in no particular order, such as
1. Red
2. Blue
3. Yellow

- **P**, the suggestions should be, in no particular order
1. Purple
2. Pink

Aradhya-B:
# Changes
- Add method to LabelService to get all course and school labels as a HashMap
- Use that method in new servlet to get all course and school labels and return as JSON
# Example Response
6 of 6 baby!
image
image

*Note: I merged #54 to this branch when I first branched to do testing but removed those changes because #54 is still outstanding at the time I'm opening this PR.*

### DIFF
--- a/src/main/java/com/google/starfish/services/LabelService.java
+++ b/src/main/java/com/google/starfish/services/LabelService.java
@@ -4,6 +4,10 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;  
 import javax.sql.DataSource;
+import java.sql.ResultSet;
+import java.util.HashMap;
+import java.util.List;
+import java.util.ArrayList;
 
 /** Enum that holds label types */
 enum Type {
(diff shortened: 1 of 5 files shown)


keelywan (approved):
Looks good!

andres-holguin (approved):
Love how you organized the school and course labels for the JSON response. This will make it super easy to handle in the front-end.

nickKelman (approved):

src/main/java/com/google/starfish/services/LabelService.java
```diff
@@ -46,6 +50,57 @@ public void insertMiscLabel(DataSource pool, String label) {
...
+          while (schoolResults.next()) {
+            String school = schoolResults.getString("title");
+            schools.add(school);
+          }
+          while (courseResults.next()) {
+            String course = courseResults.getString("title");
+            courses.add(course);
+          }
+          organizedLabels.put("schools", schools.toArray(new String[0]));
+          organizedLabels.put("courses", courses.toArray(new String[0]));
```
cftorres: since you're doing the same thing for both schools and courses, there's probably a way you can extract this behavior out into a helper function and condense the logic in this try branch a bit more

src/main/java/com/google/starfish/services/LabelService.java
```diff
@@ -46,6 +50,57 @@ public void insertMiscLabel(DataSource pool, String label) {
...
 
+  /** Returns a map of all school and course labels in the database */
+  public HashMap<String, String[]> getAllSchoolAndCourseLabels(DataSource pool) {
+    HashMap<String, String[]> organizedLabels = new HashMap<>();
+    try (Connection conn = pool.getConnection()) {
+      List<String> schools = new ArrayList<>();
+      List<String> courses = new ArrayList<>();
+      try {
+        conn.setAutoCommit(false);
+        String stmt1 =
```
cftorres: nit: since there's more than one statement here, might be more readable if you called them something a little more descriptive like `schoolsStmt` and `coursesStmt` than just using numbers to distinguish.

cftorres (approved):
Nice thinking to make this data more efficient to use on the front-end!

src/main/java/com/google/starfish/services/LabelService.java
Aradhya-B: This is a good point, I'll make a function that takes a ResultSet and returns an Array of everything in the ResultSet; that should make this more readable.

src/main/java/com/google/starfish/services/LabelService.java
Aradhya-B: Will do, nice catch!

PR #58, opened 2020-07-21
Add Search filters to Notes Grid

### LINKED ISSUE #20: Filter search by extra labels
Allow the user to filter their search result by the default labels on the notes.
The _n_ most common labels will be displayed as suggestions. Users can choose to toggle on or off the filters for these suggested labels. The user can also type in their own filter into a text box, which will generate suggestions similarly to how labels are suggested when a label is being added to a notes object.
### Steps
- [x] Determine _n_
- [x] Display the n most common labels in the search result as a suggestions for the user to filter by
- [x] All labels in the search result can be added to the filter by typing into a text box. Text box will generate suggestions based on what the user has typed in.

andres-holguin:
# Search Results Filter
This PR should close #20
The ability to filter search results has been added.
Starfish-Filter-List-V1

## Main Features
- A new FilterList component that is inside the NoteGrid component.
  - Has checkboxes with predetermined set of Common MiscLabels to filter by.
  - Updates a combobox where user can enter their own filters.
- A helper component: MockSearchButton.vue
  - Acts like the SearchBar--will emit an array of sample NoteObjects which can be used for testing the UI without a backend.

## Todo:
- Enhance Styling
- Modify FilterList to fetch suggestedLabels when School or Course changes.
- Create backend servlet that serves common MiscLabels given School/Course
- Modify SearchBar to emit current query's School and Course


src/main/webapp/components/FilterList.vue
```diff
@@ -0,0 +1,66 @@
...
+        return this.value;
+      },
+      set: function(filters) {
+        this.$emit('input', filters)
+      }
+    },
+  },
+  methods: {
+    fetchSuggestedLabels: function() {
+      let url = new URL("/common-labels", window.location.href);
```
keelywan: Is this the servlet for retrieving all the school and course labels? Or is this something else?

keelywan (approved):
This looks really good!

src/main/webapp/components/FilterList.vue
Aradhya-B: If it is a call to the servlet for retrieving all the school and course labels, I named the path `/get-school-and-course-labels` but we could also rename it on #57 if needed.

andres-holguin: Oops, forgot a TODO. This is a servlet that does not exist yet haha. It should serve some of the most common labels for each school/course.
```suggestion
      let url = new URL("/common-labels", window.location.href); // TODO: Create the servlet and rename "/common-labels"
```

src/main/webapp/components/NoteGridCollection.vue
```diff
@@ -18,11 +18,31 @@ module.exports = {
...
+    filters: Array,
+  },
+  computed: {
+    activeNotes: function() {
+      // Filter the notes that get shown based on filter provided by parent.
+      if (this.filters && this.filters.length) {
+        return this.notes.filter(this.noteFilter);
+      } else {
+        return this.notes;
+      }
```
Aradhya-B: nit: Don't need `else` if you're returning from inside the `if`.
```suggestion
      if (this.filters && this.filters.length) {
        return this.notes.filter(this.noteFilter);
      }
      return this.notes;
```

src/main/webapp/test.html
```diff
@@ -1,30 +1,62 @@
...
+    <title>Starfish</title>
+    <link href="https://fonts.googleapis.com/css2?family=Roboto&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css">
+    <script src="https://code.jquery.com/jquery-3.5.1.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.11.0/umd/popper.min.js"></script>
+    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.5.0/js/bootstrap.min.js"></script>
+    <link rel="stylesheet" href="edit-profile.css">
+    <script src="edit-profile.js"></script>
+    <link rel="stylesheet" href="styles/landingpage.css">
+    <!-- <script src="script.js"></script> -->
```
Aradhya-B, line 14: Do we still need this commented out code?

src/main/webapp/components/FilterList.vue
```diff
@@ -0,0 +1,66 @@
...
+      url.searchParams.set('school', this.school);
+      url.searchParams.set('course', this.course);
+
+      fetch(url)
+        .then(response => response.json())
+        .then(result => {
+          this.suggestedFilters = result;
+        })
```
Aradhya-B, line 55: If this is a call to the backend to get all the school and course labels, the way I ended up designing the API was that you'd get back an object where the two keys are `"schools"` and `"courses"` and their values are String arrays of labels of those type. So you could either just combine those two arrays and make that the `suggestedFilters` or you could keep two variables `suggestedSchoolFilters` and `suggestedCourseFilters` for the two arrays; I'm not sure which of these options is more useful to you on the frontend. Feel free to add a review comment on #57 if you think that API should respond differently than it does.

Aradhya-B (changes requested):
Looks really good, great work!!

src/main/webapp/components/FilterList.vue
Aradhya-B: ohhhh gotcha, then ignore my comment on aligning with the servlet in #57!

Aradhya-B (approved):

src/main/webapp/components/FilterList.vue
Aradhya-B: I can go ahead and create the `/common-labels` servlet today, I'll add it to the project board.

Aradhya-B: Now that I think about it though, I'm not sure we'd be able to create a `/common-labels` servlet without making a modification to the schema to track the number of times any given label has been searched for. Like each label is only 1 row in the Labels table at this point so I think to get the 10 most common for example, we'd need to add a column on the Labels table `numTimesSearched` or something that is basically just a counter that goes up by 1 everytime that label is searched for. Then we could simply query for labels ordered descending by `numTimesSearched`.
I could definitely work on this but I'm not sure if this is something that we want to include in our scope for this week or whether this should be considered a moonshot. If it isn't something we want to get done immediately, then it'd be good to make changes here to reflect that and assume we don't have a `/common-labels` path.

src/main/webapp/test.html
andres-holguin: No, that's my bad, this file shouldn't have even been committed. It's basically just a sandbox for debugging, shouldn't be on master.

src/main/webapp/components/NoteGridCollection.vue
andres-holguin: Thanks for the suggestion :)

andres-holguin:
> We found a Contributor License Agreement for you (the sender of this pull request), but were unable to find agreements for all the commit author(s) or Co-authors. If you authored these, maybe you used a different email address in the git commits than was used to sign the CLA ([login here](https://cla.developers.google.com/) to double check)? If these were authored by someone else, then they will need to sign a CLA as well, and confirm that they're okay with these being contributed to Google.
> In order to pass this check, please resolve this problem and then comment `@googlebot I fixed it.`. If the bot doesn't comment, it means it doesn't think anything has changed.
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fgoogleinterns%2Fstep102-2020%2Fpull%2F58) for more info**.

@Aradhya-B would you mind signing the CLA for your aradhyabansal2000@gmail.com email? If not I can probably just force push to remove that last commit.

src/main/webapp/components/FilterList.vue
andres-holguin: I was thinking the most common labels assigned to the notes on the DB, not searched for. So something like
```
SELECT label FROM
  miscLabels INNER JOIN (
    SELECT noteID FROM notes
    WHERE course="course" AND school="school"
  ) 
GROUP BY label
SORT BY count(label) DESC
LIMIT 10
 ```
I'm sure this will require a lot of tweaking, but no need to change the schema, I think.

nickKelman (approved):

PR #59, opened 2020-07-21
This adds methods to the UserService class to track increases in points based on an event

### LINKED ISSUE #21: Track prestige
Keep track of the aggregation of "rating" scores for all of the notes a user has submitted, and display this value on their profile page.
- [x] Compute prestige score based "rating" score of all user-submitted notes.
- [x] Store prestige score in database with the user profile
- [x] Display the prestige score on hover of the author name

Aradhya-B:
# Description
Depending on the event occurred, a user's points will either need to increase or decrease. Events are kept in an enum and it is easy to add new events to increase/decrease points. The steps one would have to take to add a new event would be:
1. Add event to enum
2. Add final variable constant to indicate the point modifier for that event
3. Add a case to the switch statement in `getPointsModifier`
4. Create a new method which calls `increasePoints` with the new event

For now, the only two events are Favorite and Download: the Download event is given more weight than the Favorite event but the values of the points modifiers can easily be changed if needed.

# Testing
## Before and after calling`increasePointsOnFavorite` and `increasePointsOnDownload` for user "118020203765270061430"
image

### DIFF
--- a/src/main/java/com/google/starfish/services/UserService.java
+++ b/src/main/java/com/google/starfish/services/UserService.java
@@ -3,15 +3,86 @@ package com.google.starfish.services;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;  
-import java.sql.ResultSet;  
 import javax.sql.DataSource;
+
+enum Event {
+  FAVORITE, DOWNLOAD
+}
 
 /**
  * Service class for Users Table 
  */
 public class UserService extends TableService {
 
+  private String USERS = Table.USERS.getSqlTable();
+
+  // Points modifiers for different events
+  private final long FAVORITE_POINTS_MODIFIER = 3;
+  private final long DOWNLOAD_POINTS_MODIFIER = 5;
+
   public UserService() {
     super(Table.USERS);
+  }
+
+  /** Increase user points when a user's note gets favorited */
+  public boolean increasePointsOnFavorite(DataSource pool, String userId) {
+    return increasePoints(pool, userId, Event.FAVORITE);
+  }
+
+  /** Increase user points when a user's note gets downloaded */
+  public boolean increasePointsOnDownload(DataSource pool, String userId) {
+    return increasePoints(pool, userId, Event.DOWNLOAD);
+  }
+
+  /** Variably increase user points based on event */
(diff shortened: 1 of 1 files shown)


keelywan (approved):
Looks great!

src/main/java/com/google/starfish/services/UserService.java
```diff
@@ -3,15 +3,86 @@
 import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;  
-import java.sql.ResultSet;  
 import javax.sql.DataSource;
 
+enum Event {
+  FAVORITE, DOWNLOAD
+}
+
```
andres-holguin, line 11: So scalable :0

andres-holguin (approved):
Very nice way to set this up for future events. Question: We still need a servlet to call for actually triggering these events, right?

Aradhya-B:
> Very nice way to set this up for future events. Question: We still need a servlet to call for actually triggering these events, right?

Yep, these methods would be called from servlets, so for example, we'll have a servlet to handle favoriting notes and that servlet would call `increasePointsOnFavorite` passing the `authorId`.

nickKelman (approved):

src/main/java/com/google/starfish/services/UserService.java
```diff
@@ -3,15 +3,86 @@
...
+  /** Increase user points when a user's note gets downloaded */
+  public boolean increasePointsOnDownload(DataSource pool, String userId) {
+    return increasePoints(pool, userId, Event.DOWNLOAD);
+  }
+
+  /** Variably increase user points based on event */
+  private boolean increasePoints(DataSource pool, String userId, Event event) {
+    if (userId == null) return false;
+    long pointsModifier = getPointsModifier(event);
```
cftorres: nit: prefer moving this above line 49 below to declare it only when it's needed (in case an error is thrown before even reaching that nested try, you'd be needlessly declaring this variable)

cftorres (approved):
Awesome!

src/main/java/com/google/starfish/services/UserService.java
Aradhya-B: Ahh that makes sense, nice catch!

PR #63, opened 2020-07-22
Integrate chrome extension with webapp servlets

keelywan:
## Popup 
- Added inputs for the school and course. These will be automatically set to 'Unaffiliated' if the user does not enter anything.
- The Open Browser button now works - it will open the Starfish webapp in a new tab (however see TODO at bottom)
updated-popup

## User Authentication 
- Once the chrome extension retrieves the OAuth tokens, it makes a request to the `user-registration` servlet with the `idToken` param. 
- When the user logs out, the chrome extension will revoke the `accessToken` and make a request to the `user-logout` servlet which will remove the cookies that were set when the user was registered. 

## Posting a Note
- The `UploadGNoteServlet` handles posting the note to the database.
- After the Google Doc is created, the PDF export link is retrieved and then the title, school, course, pdfSource, and sourceUrl (which is the link to the Google Doc) will be sent as parameters to the `upload-gnote` servlet.
- After the note is added to the database, the chrome extension will send a batch update request to replace all the "[DATE]" fields in the template with the current date.
- After the batch update request, the chrome extension will send another request to create global permissions. This means that anyone with the link to the Google Doc can view it.
- Once the request is completed, the Google Doc will open in a new tab

TODO: One the webapp is deployed, `WEBAPP_URL` will need to be updated. It is currently set to my local development URL.

### DIFF
--- a/chrome-extension/popup.js
+++ b/chrome-extension/popup.js
@@ -256,15 +256,16 @@ function logoutUserOnWebapp() {
 function setAccountInfo() {
   if(loggedIn) {
     gapi.client.request({
-      path: USER_INFO_URL + accessToken
-    }).then(response => {
-      let userInfo = JSON.parse(response.body);
-      document.getElementById('logout-btn').style.display = 'inline';
-      document.getElementById('email').textContent = userInfo.email;
-      document.getElementById('user-info').style.display = 'block';
-      document.getElementById('login-btn').style.display = 'none';
-      document.getElementById('generate-note-btn').disabled = false;
-    });
+      path: WEBAPP_URL + '/user-registration',
+      method: 'GET'
+    }).then(response => response.json())
+      .then(userInfo => {
+        document.getElementById('logout-btn').style.display = 'inline';
+        document.getElementById('email').textContent = userInfo.email;
+        document.getElementById('user-info').style.display = 'block';
+        document.getElementById('login-btn').style.display = 'none';
+        document.getElementById('generate-note-btn').disabled = false;
+      });
   } else {
     document.getElementById('logout-btn').style.display = 'none';
     document.getElementById('email').textContent = '';
--- a/chrome-extension/popup.js
+++ b/chrome-extension/popup.js
@@ -103,7 +103,6 @@ function getTokenEndpoint(url) {
 
 /** Sends request to webapp's user registration servlet. */
 function registerUserOnWebapp(idToken) {
-  // TODO: Change URL to deployed website URL
   fetch(WEBAPP_URL + '/user-registration?idToken=' + idToken, {
     method: 'POST'
   }).then(() => {
@@ -124,28 +123,48 @@ function getDate() {
  * with the newly created Google Doc.
  */
 function generateNote() {
-  let loadingIcon = document.getElementById('loading');
-  loadingIcon.style.display = 'flex';
-  let docName = document.getElementById('doc-name-input').value.trim();
-  if(docName === '') {
-    docName = 'gNote ' + getDate();
-  }
+  postNoteToDatabase();
+  // let loadingIcon = document.getElementById('loading');
+  // loadingIcon.style.display = 'flex';
+  // let docName = document.getElementById('doc-name-input').value.trim();
+  // if(docName === '') {
+  //   docName = 'gNote ' + getDate();
+  // }
 
-  gapi.client.request({
-    path: COPY_FILE_URL,
+  // gapi.client.request({
+  //   path: COPY_FILE_URL,
+  //   method: 'POST',
+  //   params: {fileId: NOTES_TEMPLATE_DOC_ID},
+  //   body: {
+  //     name: docName,
+  //   }
+  // }).then(response => {
+  //   // TODO: Make POST request to upload notes servlet
+  //   const gNoteURL = GOOGLE_DOC_URL + response.result.id;
+  //   loadingIcon.style.display = 'none';
+  //   chrome.tabs.create({ url: gNoteURL });
+  // }).catch(error => {
+  //   loadingIcon.style.display = 'none';
+  //   console.log('Error:', error);
+  // })
+}
+
+/**
+ * Makes a request to the webapp's upload gnote servlet to add 
+ * the note to the database.
+ */
+function postNoteToDatabase() {
+  let payload = {
+    title: 'Title',
+    school: 'NYU',
+    course: 'CS4410',
+    sourceUrl: 'sourceurl',
+    pdfSource: 'pdfSource'
+  }
+  let data = new URLSearchParams(payload);
+  fetch(WEBAPP_URL + '/upload-gnote', {
     method: 'POST',
-    params: {fileId: NOTES_TEMPLATE_DOC_ID},
-    body: {
-      name: docName,
-    }
-  }).then(response => {
-    // TODO: Make POST request to upload notes servlet
-    const gNoteURL = GOOGLE_DOC_URL + response.result.id;
-    loadingIcon.style.display = 'none';
-    chrome.tabs.create({ url: gNoteURL });
-  }).catch(error => {
-    loadingIcon.style.display = 'none';
-    console.log('Error:', error);
+    body: data
   })
 }
 
--- a/chrome-extension/popup.js
+++ b/chrome-extension/popup.js
@@ -124,7 +124,7 @@ function getDate() {
  */
 function setLoadingIcon(show) {
   const loadingIcon = document.getElementById('loading');
-  loadingIcon.style.display = (show ? 'flex' : 'none');
+  loadingIcon.style.display = show ? 'flex' : 'none';
 }
 
 /**
--- a/chrome-extension/popup.js
+++ b/chrome-extension/popup.js
@@ -134,10 +134,7 @@ function setLoadingIcon(show) {
  */
 function generateNote() {
   setLoadingIcon(true);
-  let docName = document.getElementById('doc-name-input').value.trim();
-  if(docName === '') {
-    docName = 'gNote ' + getDate();
-  }
+  const [docName, school, course] = retrieveParams();
 
   gapi.client.request({
     path: COPY_FILE_URL,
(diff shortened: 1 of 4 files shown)


chrome-extension/popup.js
```diff
@@ -7,6 +7,8 @@ const COPY_FILE_URL = 'https://www.googleapis.com/drive/v3/files/fileId/copy';
 const GOOGLE_DOC_URL = 'https://docs.google.com/document/d/';
 const REVOKE_TOKEN_URL = 'https://accounts.google.com/o/oauth2/revoke?token=';
 const USER_INFO_URL = 'https://www.googleapis.com/oauth2/v1/userinfo?access_token=';
+// TODO: Change WEBAPP_URL from local URL to deployed website URL
+const WEBAPP_URL = 'https://8080-cacf7a03-5e11-4b90-94f2-e81659d32917.us-east1.cloudshell.dev';
```
Aradhya-B, line 11: nit: I think this is a good place to use an environment variable because the value literally changes based on whether we're in dev or prod environment. I'm not 100% sure if it works in JS like it does in Node but it might just be `process.env.WEBAPP_URL`.

chrome-extension/popup.js
```diff
@@ -103,23 +101,40 @@ function getTokenEndpoint(url) {
...
  */
 function generateNote() {
-  let loadingIcon = document.getElementById('loading');
-  loadingIcon.style.display = 'flex';
-  let docName = document.getElementById('doc-name-input').value.trim();
-  if(docName === '') {
-    docName = 'gNote ' + getDate();
-  }
+  setLoadingIcon(true);
+  const [docName, school, course] = retrieveParams();
```
Aradhya-B, line 136: Great usage of destructuring!

chrome-extension/popup.js
```diff
@@ -129,28 +144,88 @@ function generateNote() {
...
+      const pdfSource = response.result.exportLinks['application/pdf'];
+      const payload = {
+        title: title,
+        school: school,
+        course: course,
+        sourceUrl: docUrl,
+        pdfSource: pdfSource
+      }
+      const noteData = new URLSearchParams(payload);
+      postNoteToDatabase(noteData)
```
Aradhya-B, line 179: nit: missing semicolon

chrome-extension/popup.js
```diff
@@ -129,28 +144,88 @@ function generateNote() {
...
+        .then(() => {
+          setLoadingIcon(false);
+          chrome.tabs.create({ url: docUrl })
```
Aradhya-B: nit: missing semicolon

chrome-extension/popup.js
```diff
@@ -129,28 +144,88 @@ function generateNote() {
...
+  getPDFLink(docId)
+    .then(response => {
+      const pdfSource = response.result.exportLinks['application/pdf'];
+      const payload = {
+        title: title,
+        school: school,
+        course: course,
+        sourceUrl: docUrl,
+        pdfSource: pdfSource
+      }
```
Aradhya-B: nit: missing semicolon

chrome-extension/popup.js
```diff
@@ -129,28 +144,88 @@ function generateNote() {
...
+
+/**
+ * Makes a request to the webapp's upload gnote servlet to add 
+ * the note to the database.
+ */
+function postNoteToDatabase(noteData) {
+  return fetch(WEBAPP_URL + '/upload-gnote', {
+    method: 'POST',
+    body: noteData
+  })
```
Aradhya-B: nit: missing semicolon

chrome-extension/popup.js
```diff
@@ -161,7 +236,7 @@ function logout() {
 function setAccountInfo() {
   if(loggedIn) {
     gapi.client.request({
-      path: USER_INFO_URL + accesToken
+      path: USER_INFO_URL + accessToken
     }).then(response => {
```
Aradhya-B: Should we be requesting this information from our own `user-registration` servlet on a GET request now that we have that?

src/main/java/com/google/starfish/servlets/UploadGNoteServlet.java
```diff
@@ -0,0 +1,122 @@
...
+   **/
+  private boolean validateUser(HttpServletRequest req) {
+    Cookie[] cookies = req.getCookies();
+    String sessionId = null;
+    for (Cookie cookie : cookies) {
+      if (COOKIE_NAME.equals(cookie.getName())) {
+        sessionId = cookie.getValue();
+        break;
+      }
+    }
```
Aradhya-B, line 109: Not a huge deal, but possible todo is to make this a static method of a `ValidateUser` class or something because we're using this exact code in more than one servlet.

src/main/java/com/google/starfish/servlets/UploadGNoteServlet.java
```diff
@@ -0,0 +1,122 @@
...
+
+    HttpSession activeSession = request.getSession(false);
+    String userId = (String) activeSession.getAttribute("user_id");
+
+    String title = request.getParameter("title");
+    String school = request.getParameter("school").toLowerCase();
+    String course = request.getParameter("course").toLowerCase();
+    String sourceUrl = request.getParameter("sourceUrl");
+    String pdfSource = request.getParameter("pdfSource");
+
```
Aradhya-B, line 48: Should we have a field in the chrome extension to add misc labels separated by commas? We already have the ability to handle this on the backend.

src/main/java/com/google/starfish/servlets/UploadGNoteServlet.java
```diff
@@ -0,0 +1,122 @@
...
+              + "?,"
+              + "?,"
+              + "?,"
+              + "?,"
+              + "?,"
+              + "?,"
+              + "?,"
+              + "0 ); ";
+
+      try (PreparedStatement noteStmt = conn.prepareStatement(stmt, new String[] {"id"})) {
```
Aradhya-B: nit: You don't need the second param in `prepareStatement`, that was only added in the other servlet because I needed to get the `id` of the inserted row back by calling `generateKeys`.

Aradhya-B (approved):
Really great work!! Just added a couple nit picky comments but overall LGTM :).

andres-holguin (approved):

chrome-extension/popup.js
keelywan: oops my bad I keep forgetting semicolons

chrome-extension/popup.js
keelywan: You're right! I'll change that

src/main/java/com/google/starfish/servlets/UploadGNoteServlet.java
keelywan: The reason I didn't include it was because I didn't want to overwhelm the user with inputs in the chrome extension (since it's supposed to be an quick way to create a note), but if you think it'd be better to include it, I can add it.

src/main/java/com/google/starfish/servlets/UploadGNoteServlet.java
keelywan: Ohh I see, I'll remove that!

chrome-extension/popup.js
keelywan: Wait does there need to be a semicolon here if there is a `.then()`?

chrome-extension/popup.js
keelywan: I'm not sure what you mean about the environment variable, are you saying to set `WEBAPP_URL` to `process.env.WEBAPP_URL`?

src/main/java/com/google/starfish/servlets/UploadGNoteServlet.java
keelywan: Yeah that makes sense, maybe we can add this in a future PR.

chrome-extension/popup.js
Aradhya-B: Ohhh no we don't, I didn't realize this `.then()` was a nested `.then()` haha

chrome-extension/popup.js
Aradhya-B: Yep, then we can set the webapp url in the environment variables in appengine-web.xml for prod and in the env_variables file for dev

src/main/java/com/google/starfish/servlets/UploadGNoteServlet.java
Aradhya-B: Ah I see, no that makes a lot of sense and I think you made the right decision. Was just wondering cause we could technically handle it but this is more of a UX thing.

chrome-extension/popup.js
keelywan: I'm not sure if that would work actually because when the chrome extension is going to be run separately from our webapp once it's on the app store.

Aradhya-B: Ohhh I see, yeah that makes sense. Hmmm then I don't think there's a way to set the url dynamically from the chrome extension based on whether we're in prod or dev. I guess it's not too big of a deal to hardcode it for now then

chrome-extension/popup.js
```diff
@@ -129,28 +143,122 @@ function generateNote() {
...
-    loadingIcon.style.display = 'none';
+    setLoadingIcon(false);
     console.log('Error:', error);
-  })
+  });
+}
+
+/** Retrieves input values and sets them if they are blank */
+function retrieveParams() {
+  let docName = document.getElementById('doc-name-input').value.trim();
```
nickKelman: Can simplify logic like this in the following way:

const docName = document.getElementById('doc-name-input').value.trim() || ('gNote ' + getDate());

Applies to below fields

chrome-extension/popup.js
```diff
@@ -129,28 +143,122 @@ function generateNote() {
...
+            matchCase: true
+          }
+        },
+      }],
+    },
+  };
+  gapi.client.docs.documents.batchUpdate(updateObject)
+    .then(() => addGlobalPermissions(docId))
+    .then(() => {
+      setLoadingIcon(false);
```
nickKelman: nit:

I assumed you want the loading icon to stop displaying even if an error occurs during batch update. This logic won't run on error, so I think you want the loading Icon in a finally (Or something of the like) statement that gets run on success or failure.

nickKelman (approved):
Nice work.

cftorres (approved):
nice use of helper methods!

PR #65, opened 2020-07-23
Implement downloading logic on NotePreview

keelywan:
## NotePreview
- Added ability to download uploaded PDF when the download badge on `NotePreview` is clicked 
  - Downloading a note will increase the number of downloads on the note itself
  - Downloading a note will also increase the number of points on the user who posted the note

Also fixed small bug in search servlet.


src/main/java/com/google/starfish/servlets/DownloadNoteServlet.java
```diff
@@ -0,0 +1,39 @@
...
+  @Override
+  public void doPost(HttpServletRequest req, HttpServletResponse res) throws ServletException, IOException {
+    Long noteId = Long.valueOf(req.getParameter("note_id"));
+    DataSource pool = (DataSource) req.getServletContext().getAttribute("my-pool");  
+
+    try (Connection conn = pool.getConnection()) {
+      noteService.incrementDownloadsByNoteId(pool, noteId);
+      Note downloadedNote = noteService.getNoteByNoteId(pool, noteId);
+      String userId = downloadedNote.getAuthorId();
+      userService.increasePointsOnDownload(pool, userId);
```
Aradhya-B: nit: would prefer this variable to be called `authorId` to differentiate between "this user" and the "user who uploaded the note".

Aradhya-B (approved):
Awesome work!!

src/main/java/com/google/starfish/servlets/DownloadNoteServlet.java
keelywan: Oh right that's definitely more intuitive, I'll make that change!

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
```diff
@@ -53,6 +53,8 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       response.setStatus(HttpServletResponse.SC_NOT_ACCEPTABLE);
       return;
     }
+    
+    blobKey = "/serve-notes?key=" + blobKey;
```
andres-holguin: Since this is no longer a blobKey, but the sourceUrl, could we rename it that and trickle down the change?

andres-holguin (approved):
Awesome stuff!

src/main/java/com/google/starfish/servlets/HandleNotesServlet.java
keelywan: Yep I'll make that change!

cftorres (approved):

PR #70, opened 2020-07-24
Implement favoriting logic on NotePreview

keelywan:
## Navbar 
- Added a `getUserInfo` method that makes a request to the user-registration servlet to get the user object
- The dropdown now displays the user's display name and points instead of being hardcoded
updated-dropdown

## NotePreview
- When the note is favorited by the user, the star icon will be yellow instead of grey
- Also fixed the bug so that the badges will display the number zero in the blue circle, instead of leaving it empty

favorite-star

- Note: The number of favorites is not being updated in real time. For example, if a user unfavorites a note, the number displayed in the badge will not decrease.

### doGet 
The doGet takes `note_id` as a request parameter and will return true if the user has favorited this note and false if they haven't or there isn't a user logged in.

### doPost 
The doPost takes `note_id` as a request parameter and will add a new row to the favorite_notes table in the database. It will also use the UserService to increment the number of points for the author of the note. 

### doDelete
The doDelete takes `note_id` as a request parameter and will delete the row in favorite_notes table containing the `note_id` and `user_id`. It will also use the UserService to decrement the number of points for the author of the note. 

## UserService 
- Added an `UNFAVORITE` event and a `decreasePointsOnUnfavorite` method which will be called whenever a user unfavorites a note


src/main/java/com/google/starfish/servlets/FavoriteNoteServlet.java
```diff
@@ -0,0 +1,135 @@
...
+
+    HttpSession activeSession = req.getSession(false);
+    String userId = (String) activeSession.getAttribute("user_id");
+    Long noteId = Long.valueOf(req.getParameter("note_id"));
+    DataSource pool = (DataSource) req.getServletContext().getAttribute("my-pool");
+
+    try (Connection conn = pool.getConnection()) {
+      favoriteNoteService.deleteRowByCompoundId(pool, noteId, userId);
+      Note downloadedNote = noteService.getNoteByNoteId(pool, noteId);
+      String authorId = downloadedNote.getAuthorId();
```
Aradhya-B: nit: don't think `downloadedNote` is the right name here. Prefer if this was just 1 line like `String authorId = noteService.getNoteByNoteId(pool, noteId).getAuthorId();`.

Aradhya-B (approved):
Amazing work!!!

src/main/java/com/google/starfish/servlets/FavoriteNoteServlet.java
keelywan: My bad I think I might've copy pasted code from the download-logic. I'll change that!

src/main/webapp/components/NotePreview.vue
```diff
@@ -77,6 +82,39 @@
...
+            this.favorited = isFavorited;
+            const method = this.favorited ? 'DELETE' : 'POST';
+            fetch('/favorite-note?note_id=' + this.id, {
+              method: method
+            }).then(response => {
+              if(response.status === 403) alert('Please sign in to favorite this note.');
+              else this.favorited = !this.favorited;
+            })
+          })
+      },
```
andres-holguin, line 93: Looks like NoteGridCard.vue is still using the old toggleFavorite function. Could you add that change to this PR?
Edit: Since the functions should be nearly identical, maybe this should be a method in a common parent that gets passed down to NotePreview and NoteCard as a prop? Alternatively, remove the ability to favorite from the card altogether, since the user shouldn't be allowed to favorite a note they haven't previewed.

keelywan: I'm not sure where you see that NoteGridCard.vue is using the old toggleFavorite function. Is it in this PR?

andres-holguin: No, but I made these changes in #73 , so should be good to go :)

andres-holguin (approved):

src/main/webapp/components/NotePreview.vue
keelywan: Ok thanks!

src/main/java/com/google/starfish/servlets/FavoriteNoteServlet.java
```diff
@@ -0,0 +1,133 @@
...
+    } catch (SQLException ex) {
+      System.err.print(ex);
+    }
+  }
+
+  /**
+   * Validates the user using the request's session and cookies. If there is a
+   * valid user logged in, returns true. Otherwise, returns false.
+   **/
+  private boolean validateUser(HttpServletRequest req) {
```
nickKelman, line 112: If this logic is spread across each servlet, could you consolidate the logic and share it through a utility service?

nickKelman (approved):

src/main/java/com/google/starfish/servlets/FavoriteNoteServlet.java
keelywan: Yep, I'll do that!

PR #72, opened 2020-07-24
This adds a servlet to return trending notes in a given timespan and the associated service methods to determine which notes are trending

### LINKED ISSUE #19: Rank notes
Each notes object will be assigned a score based on:
- Number of Downloads (low-weight)
- Number of Favorites (high-weight)

Notes can then be sorted based on this "rating" score.
### Steps
- [x] Keep track of number of Downloads and Favorites in Notes object
- [x] Compute a "rating" score based on number of Downloads and Favorites
- [x] Sort by this "rating" score.
- [ ] Sort by change in score over the last X weeks
- [ ] Safety Check: Front and Backend

### LINKED ISSUE #68: Get Trending Notes
Sub-issue for #19 
- [ ] Incorporate into Front-end

Aradhya-B:
# Description
The Starfish platform needs a way to rank notes: notes will be ranked according to how many times they've been favorited and the most favorited notes will be the most "trending". This PR will give users a way to see which notes have been most favorited today, this week, this month, or all time, helping them find the most popular notes as quickly as possible. It also changes the search servlet to a 2 dimensional array where each element of the outer array is another array that holds a note in index 0, and the number of times its been favorited in the given timespan in index 1. The array is sorted as it is querried from sql by most trending notes first.

The trending-notes servlet can be used to create a general trending page that shows the most trending notes, irrespective of school or course, and the search servlet is used to make more refined queries that still rank by trending, but are filtered by a school and/or course
# Additions
- Add a `date` column to the `favorite_notes` table to keep track of when a note gets favorited
- Add methods to the `FavoriteNoteService` class to return trending notes today, this week, this month, and all time
- Add a servlet that takes a query string param, `timespan`, based on which the servlet returns a 2D array sorted by most popular note first in the given timespan
- Change search servlet to take additional query string param, `timespan`, and use to return 2D array sorted by most popular note first in the given timespan and filtered by school and/or course
# Example Response From Search With No Passed Params (Default All Notes, All-Time Recency)
image

### DIFF
--- a/src/main/java/com/google/starfish/services/FavoriteNoteService.java
+++ b/src/main/java/com/google/starfish/services/FavoriteNoteService.java
@@ -11,18 +11,18 @@ import java.util.ArrayList;
 import java.util.Calendar;
 import com.google.starfish.models.Note;
 
-/** Enum to hold possible recency to get trending notes */
-enum Recency {
-  TODAY, 
-  THIS_WEEK, 
-  THIS_MONTH, 
-  ALL_TIME;
-}
-
 /**
  * Service class for FavoriteNotes Table
  */
 public class FavoriteNoteService {
+
+  /** Enum to hold possible recency to get trending notes */
+  public enum Recency {
+    TODAY, 
+    THIS_WEEK, 
+    THIS_MONTH, 
+    ALL_TIME;
+  }
 
   private String FAVORITE_NOTES = Table.FAVORITE_NOTES.getSqlTable();
   private String NOTES = Table.NOTES.getSqlTable();
@@ -128,48 +128,8 @@ public class FavoriteNoteService {
     }
   }
 
-  /** Gets trending notes today */
-  public Object[][] getTrendingNotesToday(DataSource pool) throws SQLException {
-    return getTrendingNotes(pool, Recency.TODAY);
-  }
-
-  /** Gets trending notes this week */
-  public Object[][] getTrendingNotesThisWeek(DataSource pool) throws SQLException {
-    return getTrendingNotes(pool, Recency.THIS_WEEK);
-  }
-
-  /** Gets trending notes this month */
-  public Object[][] getTrendingNotesThisMonth(DataSource pool) throws SQLException {
-    return getTrendingNotes(pool, Recency.THIS_MONTH);
-  }
-
-  /** Gets all-time trending notes */
-  public Object[][] getTrendingNotesAllTime(DataSource pool) throws SQLException {
-    return getTrendingNotes(pool, Recency.ALL_TIME);
-  }
-
-  /** Gets trending notes today filtered by school or course */
-  public Object[][] getTrendingNotesTodayBySchoolOrCourse(DataSource pool, String school, String course) throws SQLException {
-    return getTrendingNotesBySchoolOrCourse(pool, Recency.TODAY, school, course);
-  }
-
-  /** Gets trending notes this week filtered by school or course */
-  public Object[][] getTrendingNotesThisWeekBySchoolOrCourse(DataSource pool, String school, String course) throws SQLException {
-    return getTrendingNotesBySchoolOrCourse(pool, Recency.THIS_WEEK, school, course);
-  }
-
-  /** Gets trending notes this month filtered by school or course */
-  public Object[][] getTrendingNotesThisMonthBySchoolOrCourse(DataSource pool, String school, String course) throws SQLException {
-    return getTrendingNotesBySchoolOrCourse(pool, Recency.THIS_MONTH, school, course);
-  }
-
-  /** Gets all-time trending notes filtered by school or course */
-  public Object[][] getTrendingNotesAllTimeBySchoolOrCourse(DataSource pool, String school, String course) throws SQLException {
-    return getTrendingNotesBySchoolOrCourse(pool, Recency.ALL_TIME, school, course);
-  }
-
   /** Gets trending notes based on number of favorites in a given timespan, filtered by school and/or course */
-  private Object[][] getTrendingNotesBySchoolOrCourse(DataSource pool, Recency recency, String school, String course) throws SQLException {
+  public Object[][] getTrendingNotesBySchoolOrCourse(DataSource pool, Recency recency, String school, String course) throws SQLException {
     String filterStmt = getStatementToFilterBySchoolOrCourse(school, course);
     Date date = getDateBasedOnRecency(recency);
     List<Object> notes = new ArrayList<>();
@@ -201,26 +161,6 @@ public class FavoriteNoteService {
     return filterStmt;
   }
 
-  /** Gets trending notes based on number of favorites in a given timespan */
-  private Object[][] getTrendingNotes(DataSource pool, Recency recency) throws SQLException {
-    Date date = getDateBasedOnRecency(recency);
-    List<Object> notes = new ArrayList<>();
-    try (Connection conn = pool.getConnection()) {
-      String stmt = getTrendingNotesStatement(null);
-      try (PreparedStatement selectStmt = conn.prepareStatement(stmt)) {
-        selectStmt.setDate(1, date);
-        ResultSet rs = selectStmt.executeQuery();
-        while (rs.next()) {
-          Note thisNote = constructNoteFromSqlResult(pool, rs);
-          Long numFavoritesInTimespan = rs.getLong(NUM_FAVORITES_IN_TIMESPAN_ID);
-          notes.add(new Object[] {thisNote, numFavoritesInTimespan});
-        }
-        rs.close();
-        return notes.toArray(new Object[0][0]);
-      }
-    }
-  }
-
   /** Gets the sql query to return trending notes possibly filtered by school and/or course */
   private String getTrendingNotesStatement(String schoolAndCourseFilter) {
     String stmt = 
@@ -242,24 +182,20 @@ public class FavoriteNoteService {
     Calendar calendar = Calendar.getInstance();
     switch(recency) {
       case TODAY:
-        date = new Date(calendar.getTimeInMillis());
         break;
       case THIS_WEEK:
         calendar.add(Calendar.DAY_OF_MONTH, -7);
-        date = new Date(calendar.getTimeInMillis());
         break;
       case THIS_MONTH:
         calendar.add(Calendar.DAY_OF_MONTH, -30);
-        date = new Date(calendar.getTimeInMillis());
         break;
       case ALL_TIME:
         // No note should ever be favorited more than 100 years ago (for now...)
         calendar.add(Calendar.YEAR, -100);
-        date = new Date(calendar.getTimeInMillis());
         break;
       default:
-        date = new Date(calendar.getTimeInMillis());
     }
+    date = new Date(calendar.getTimeInMillis());
     return date;
   }
 
--- a/src/main/java/com/google/starfish/services/FavoriteNoteService.java
+++ b/src/main/java/com/google/starfish/services/FavoriteNoteService.java
@@ -18,10 +18,18 @@ public class FavoriteNoteService {
 
   /** Enum to hold possible recency to get trending notes */
   public enum Recency {
-    TODAY, 
-    THIS_WEEK, 
-    THIS_MONTH, 
-    ALL_TIME;
+    TODAY("today"),
+    THIS_WEEK("this-week"),
+    THIS_MONTH("this-month"),
+    ALL_TIME("all-time");
+
+    private String timespan;
+    Recency(String timespan) {
+      this.timespan = timespan;
+    }
+    String getTimespan() {
+      return this.timespan;
+    }
   }
 
   private String FAVORITE_NOTES = Table.FAVORITE_NOTES.getSqlTable();
--- a/src/main/java/com/google/starfish/services/FavoriteNoteService.java
+++ b/src/main/java/com/google/starfish/services/FavoriteNoteService.java
@@ -11,6 +11,13 @@ import java.util.ArrayList;
 import java.util.Calendar;
 import com.google.starfish.models.Note;
 
+enum Recency {
+  TODAY, 
+  THIS_WEEK, 
+  THIS_MONTH, 
+  ALL_TIME;
+}
+
 /**
  * Service class for FavoriteNotes Table
  */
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -115,6 +127,81 @@ public long getNumFavoritesByNoteId(DataSource pool, long noteId) throws SQLExce
...
+
+  /** Gets trending notes this month */
+  public Note[] getTrendingNotesThisMonth(DataSource pool) throws SQLException {
+    return getTrendingNotes(pool, Recency.THIS_MONTH);
+  }
+
+  /** Gets all-time trending notes */
+  public Note[] getTrendingNotesAllTime(DataSource pool) throws SQLException {
+    return getTrendingNotes(pool, Recency.ALL_TIME);
+  }
```
andres-holguin: Love the abstraction here.

andres-holguin (approved):
Really good stuff! I'm going to start testing this on the front end. However, it looks like right now all Notes get returned. Could you add a TODO for narrowing it down by school and/or course, like in the search servlet?

src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -115,6 +127,81 @@ public long getNumFavoritesByNoteId(DataSource pool, long noteId) throws SQLExce
...
+
+  /** Gets trending notes based on number of favorites in a given timespan */
+  private Note[] getTrendingNotes(DataSource pool, Recency recency) throws SQLException {
+    Date date = getDateBasedOnRecency(recency);
+    List<Note> notes = new ArrayList<>();
+    try (Connection conn = pool.getConnection()) {
+      String stmt = 
+          "SELECT * "
+        + "FROM "
```
keelywan: nit: indentation

keelywan (approved):
Looks great!

src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -115,6 +128,141 @@ public long getNumFavoritesByNoteId(DataSource pool, long noteId) throws SQLExce
...
+      }
+    }
+  }
+
+  /** Gets the sql query to return trending notes possibly filtered by school and/or course */
+  private String getTrendingNotesStatement(String schoolAndCourseFilter) {
+    String stmt = 
+        "SELECT * "
+      + "FROM " + NOTES + " AS a "
+        + "INNER JOIN (SELECT note_id, COUNT(*) AS " + NUM_FAVORITES_IN_TIMESPAN_ID + " "
```
keelywan: nit: indentation

keelywan (approved):
LGTM!

andres-holguin (approved):
LGTM! Thank you for making those non-trivial changes to the search servlet.

src/main/java/com/google/starfish/services/FavoriteNoteService.java
Aradhya-B: I made a change!

src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -115,6 +128,141 @@ public long getNumFavoritesByNoteId(DataSource pool, long noteId) throws SQLExce
...
+      filterStmt += "AND `school`= '" + school + "' ";
+    }
+    if (course != null && !course.isEmpty()) {
+      filterStmt += "AND `course`= '" + course + "' ";
+    }
+    return filterStmt;
+  }
+
+  /** Gets trending notes based on number of favorites in a given timespan */
+  private Object[][] getTrendingNotes(DataSource pool, Recency recency) throws SQLException {
```
cftorres: I don't think this is necessary-- you can simply modify the helper method for getting the filter statement to return an empty string if there is no school or course passed in so you can avoid repeating all of this logic.

Similarly, I also feel like it would be possible to reduce the number of methods above by simply passing null for school/course so you don't have to kind of have duplicate wrappers to get trending notes per time period

src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -115,6 +128,141 @@ public long getNumFavoritesByNoteId(DataSource pool, long noteId) throws SQLExce
...
+    return getTrendingNotes(pool, Recency.THIS_MONTH);
+  }
+
+  /** Gets all-time trending notes */
+  public Object[][] getTrendingNotesAllTime(DataSource pool) throws SQLException {
+    return getTrendingNotes(pool, Recency.ALL_TIME);
+  }
+
+  /** Gets trending notes today filtered by school or course */
+  public Object[][] getTrendingNotesTodayBySchoolOrCourse(DataSource pool, String school, String course) throws SQLException {
```
cftorres: I wonder if it's really necessary to define all of these wrappers.... it might be worth it to condense this and simply call the more generic function with the specified time period from the front-end since you already have switch logic below where you can just specify it as a parameter according to the right case

src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -115,6 +128,141 @@ public long getNumFavoritesByNoteId(DataSource pool, long noteId) throws SQLExce
...
+        break;
+      case ALL_TIME:
+        // No note should ever be favorited more than 100 years ago (for now...)
+        calendar.add(Calendar.YEAR, -100);
+        date = new Date(calendar.getTimeInMillis());
+        break;
+      default:
+        date = new Date(calendar.getTimeInMillis());
+    }
+    return date;
```
cftorres, line 207: Since you set `date` in every single case-- can you not just do that at the very end after leaving the switch block (after `calendar` has been added to when necessary)?

src/main/java/com/google/starfish/servlets/GetTrendingNotesServlet.java
```diff
@@ -0,0 +1,63 @@
...
+  public void doGet(HttpServletRequest req, HttpServletResponse res) throws ServletException, IOException {
+    DataSource pool = (DataSource) req.getServletContext().getAttribute("my-pool");  
+
+    String timespan = trimAndLowerCaseString(req.getParameter("timespan"));
+    // Default timespan is all-time
+    if (timespan == null) timespan = "all-time";
+    try {
+      Object[][] trendingNotes = null;
+      switch(timespan) {
+        case "today":
```
cftorres: looks like you re-use these below so would be safer to use an enum to keep consistent

src/main/java/com/google/starfish/servlets/GetTrendingNotesServlet.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  /** If a string is not null, trims whitespace and makes it all lower case */
+  private String trimAndLowerCaseString(String string) {
+    if (string == null || string.isEmpty()) return null;
+    return string.toLowerCase().trim();
+  }
+
+  /** Converts a notes array to JSON */
```
cftorres: nit: this doesn't have to be a notes array, can be any 2D array, right?

src/main/java/com/google/starfish/servlets/SearchServlet.java
```diff
@@ -114,9 +54,9 @@ private String trimAndLowerCaseString(String string) {
     return string.toLowerCase().trim();
   }
 
-  /** Converts an array list to JSON */
-  private String convertListToJson(List<Note> notes) {
+  /** Converts a 2D array to JSON */
+  private String convert2DArrayToJSON(Object[][] arr) {
```
cftorres: can you share this function since you use it above, rather than re-defining it?

src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -115,6 +128,141 @@ public long getNumFavoritesByNoteId(DataSource pool, long noteId) throws SQLExce
...
+        return notes.toArray(new Object[0][0]);
+      }
+    }
+  }
+
+  /** Gets the sql query to return trending notes possibly filtered by school and/or course */
+  private String getTrendingNotesStatement(String schoolAndCourseFilter) {
+    String stmt = 
+        "SELECT * "
+      + "FROM " + NOTES + " AS a "
```
nickKelman, line 178: nit: whitespce is off

src/main/java/com/google/starfish/servlets/GetTrendingNotesServlet.java
nickKelman: +1

nickKelman (approved):

src/main/java/com/google/starfish/services/FavoriteNoteService.java
Aradhya-B: You're right!! I didn't catch this at all but the null checks for school and course are already being done in the filter statement and should actually be usable as is in a single getTrendingNotes method that can take a null school or course. I'll get rid of all the wrappers and make a single method that takes school and course with the usage that if school or course are null, then there's no real filter being applied.

src/main/java/com/google/starfish/services/FavoriteNoteService.java
Aradhya-B: 100%, I think the additional wrappers from having 2 main "wrapped" functions is overkill and you're right, since the switch logic is already here it probably just makes more sense to call the method with the enum directly. I'll make those changes.

src/main/java/com/google/starfish/services/FavoriteNoteService.java
Aradhya-B: Ohhh yep you're right, no need to set the value for date in the switch statement!

src/main/java/com/google/starfish/servlets/GetTrendingNotesServlet.java
Aradhya-B: Right, I'll fix the comment.

src/main/java/com/google/starfish/servlets/SearchServlet.java
Aradhya-B: Yep, I'll make a util class!

src/main/java/com/google/starfish/services/FavoriteNoteService.java
Aradhya-B: Good catch!

src/main/java/com/google/starfish/servlets/GetTrendingNotesServlet.java
Aradhya-B: Discussed offline -- we can use the existing Recency enum in the FavoriteNoteService class to enforce consistency throughout the servlets and service!

src/main/java/com/google/starfish/services/FavoriteNoteService.java
```diff
@@ -115,6 +138,77 @@ public long getNumFavoritesByNoteId(DataSource pool, long noteId) throws SQLExce
...
+    stmt += "ORDER BY count DESC;";
+    return stmt;
+  }
+
+  /** Gets the date based on recency */
+  private Date getDateBasedOnRecency(Recency recency) {
+    Date date;
+    Calendar calendar = Calendar.getInstance();
+    switch(recency) {
+      case TODAY:
```
cftorres: I think you don't need this because if it doesn't match a specific case, it just goes to `default`

src/main/java/com/google/starfish/servlets/GetTrendingNotesServlet.java
```diff
@@ -0,0 +1,51 @@
...
+
+    String timespan = Utils.trimAndLowerCaseString(req.getParameter("timespan"));
+    Recency recency = Utils.findRecencyByString(timespan);
+    // Default recency is all-time
+    if (recency == null) recency = Recency.ALL_TIME;
+    try {
+      Object[][] trendingNotes = null;
+      switch(recency) {
+        case TODAY:
+          trendingNotes = favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, FavoriteNoteService.Recency.TODAY, null, null);
```
cftorres: do you have to prepend this with `FavoriteNoteService` since you import `Recency`?

src/main/java/com/google/starfish/servlets/SearchServlet.java
```diff
@@ -1,122 +1,54 @@
...
+    String reqCourse = Utils.trimAndLowerCaseString(req.getParameter("course"));
+    String timespan = Utils.trimAndLowerCaseString(req.getParameter("timespan"));
+    Recency recency = Utils.findRecencyByString(timespan);
+    // Default recency is all-time
+    if (recency == null) recency = Recency.ALL_TIME;
+    try {
+      Object[][] trendingNotes = null;
+      switch(recency) {
+        case TODAY:
+          trendingNotes = favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, FavoriteNoteService.Recency.TODAY, reqSchool, reqCourse);
```
cftorres: same comment here as above file (but admittedly i'm not super familiar with java so might be wrong)

src/main/java/com/google/starfish/servlets/GetTrendingNotesServlet.java
```diff
@@ -0,0 +1,51 @@
...
+          break;
+        case THIS_WEEK:
+          trendingNotes = favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, FavoriteNoteService.Recency.THIS_WEEK, null, null);
+          break;
+        case THIS_MONTH:
+          trendingNotes = favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, FavoriteNoteService.Recency.THIS_MONTH, null, null);
+          break;
+        default:
+          // Default to returning trending notes all-time
+          trendingNotes = favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, FavoriteNoteService.Recency.ALL_TIME, null, null);
```
cftorres: (Here and in file below) actually do you need this switch block anymore?  Couldn't you just say `trendingNotes = favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, recency, null, null)` and it would accordingly pass the right value for recency?

src/main/java/com/google/starfish/servlets/Utils.java
```diff
@@ -0,0 +1,29 @@
...
+public class Utils {
+
+  /** Converts a 2D array to JSON */
+  public static String convert2DArrayToJSON(Object[][] arr) {
+    Gson gson = new Gson();
+    return gson.toJson(arr);
+  }
+
+  /** Finds key of Recency enum by string value */
+  public static Recency findRecencyByString(String recency) {
```
cftorres, line 15: I think this looks fine-- I like that you extracted these out to a Utils file!

src/main/java/com/google/starfish/services/FavoriteNoteService.java
Aradhya-B: Oh you're right, I wasn't a 100% sure if I should still include it because it was technically still a "case", but then again I think I don't have the `ALL_TIME` case in other places so it makes sense not to include this here.

src/main/java/com/google/starfish/servlets/GetTrendingNotesServlet.java
Aradhya-B: Nope I don't, I must've imported after I'd already prepended everywhere so my linter didn't catch it.

src/main/java/com/google/starfish/servlets/SearchServlet.java
Aradhya-B: Nope you're right, good catch!

src/main/java/com/google/starfish/servlets/GetTrendingNotesServlet.java
Aradhya-B: woahhhh you're right haha, turns out there was even more benefit to using that recency enum everywhere!! This really simplifies the code a lot more!

src/main/java/com/google/starfish/servlets/Utils.java
Aradhya-B: Thanks! Yeah we talked about that tradeoff between complexity and consistency offline but for such small enums it really isn't too big of a deal.

PR #73, opened 2020-07-27
Prettify the UI for Search Results

andres-holguin:
Note: Since this PR pulled from the currently ongoing PRs, it seems overloaded with commits. 
# UI Improvements

Starfish-Slide-Group-V1

#### EDIT: Added card highlighting
Starfish-Slide-Group-V2

This PR brings some improvements to the UI for how Note results are displayed:

### Main features
- NoteGridCollections can be navigated as a slide group with arrows on both sides.
- Long NoteGridCard titles and school names will be truncates with ellipses (...) at the end
- The thumbnail image is now contained in a <v-img> component, which allows for easy overlays
- The icons used to represent google drive files and pdf files have been changed to use material design icons.
- EDIT: Cards will be highlighted/activated when clicked on. Only activated cards open the preview on click.

### TODO:
- Set Date parsing so certain time frames are displayed as "x days/hours ago"
- Decide how to best indicate NoteRating on NoteGridCard
- EDIT: Compute whether a note is a GNote


src/main/webapp/components/NoteGridCard.vue
```diff
@@ -1,36 +1,63 @@
...
       labels: Array,
       numDownloads: Number,
       numFavorites: Number,
-      isFavorited: Boolean
+      isFavorited: Boolean,
+    },
+    data: function() {
+      return {
+        favorited: this.isFavorited,
+        isGNote: true,
```
keelywan: Do you think that this should be a computed value instead of in data?

src/main/webapp/components/NoteGridCard.vue
```diff
@@ -1,36 +1,63 @@
...
+
+      <v-card-text>
+        <v-row no-gutters>
+          <v-col>
+            {{dateCreated}}
+          </v-col>
+          <v-col col="1">
+            <v-badge overlap class="ma-1">
+              <template v-slot:badge>{{numFavorites}}</template>
+              <v-icon :color="favorited && 'yellow'">mdi-star</v-icon>
```
keelywan: I tried running this locally, and I got an error saying that the type check failed for prop color because it expected a String but got a Boolean. Maybe you could store a color variable in data and then change it whenever `this.favorited` changes?

src/main/webapp/components/NoteGridCard.vue
```diff
@@ -42,12 +69,20 @@
     methods: {
       onClick: function() {
         this.$emit('click');
+      },
+      toggleFavorite: function() {
+        let method = this.favorited ? 'DELETE' : 'POST';
+        this.favorited = !this.favorited;
+        fetch('/favorite-note?note_id=' + this.id, {
+          method: method
+        })
```
keelywan: I think this toggleFavorite method can be deleted because the toggling function should be in NotePreview instead.

keelywan (approved):
The UI looks really good!

src/main/webapp/components/NoteGridCard.vue
andres-holguin: Makes, sense did that and added a TODO for computing it :)

src/main/webapp/components/NoteGridCard.vue
andres-holguin: Ahh good catch, thought I could do something fancy but failed x) made that change.

src/main/webapp/components/NoteGridCard.vue
andres-holguin: Good catch

andres-holguin:
@keelywan re-requesting a review for new changes to card highlighting.

Aradhya-B (approved):
LGTM, but do we still need the MockSearchButton? I think now that normal search and note posting is working, we could probably do without it.

src/main/webapp/components/MockSearchButton.vue
```diff
@@ -46,6 +49,97 @@ module.exports = {
           isFavorited: false,
           pdfSource: "/assets/notes.pdf"
         },
+        {
+          thumbnailSrc: "/assets/notes.png",
+          author: "Andres H.",
+          title: "ABC set of notes",
+          dateCreated: new Date(Date.now() - 1000 * 60 * 60 * 24 * 6).toDateString(),
```
cftorres: prefer extracting this kind of math out into a constant (e.g. `SIX_DAYS` or something like `HOURS_PER_DAY * 6` if you'll be using this same constant in different ways) so it's clear how Date is being manipulated

keelywan (approved):
Looks good! Thanks for making the changes 😄

cftorres (approved):
Looking good!

andres-holguin:
> LGTM, but do we still need the MockSearchButton? I think now that normal search and note posting is working, we could probably do without it.

I tend to run a web-server that isn't connected to the back end since it's faster for testing front-end changes.

src/main/webapp/components/MockSearchButton.vue
andres-holguin: You got it!

PR #78, opened 2020-07-27
Resolve upload notes error on deployed webapp

### LINKED ISSUE #74: Fix blobstore for deployment

keelywan:
This changes the data type of `source_url` and `pdf_source` from `VARCHAR(255)` to `VARCHAR(1000)`. This is because when we deployed the app, the generated blobKey for uploaded files ended up being much larger than expected and did not fit in the table which led to a MysqlDataTruncation Exception. We fixed this by increasing the size from 255 to 1000.


Aradhya-B (approved):

andres-holguin (approved):
Does this mean we can deploy our last demo? Edit: Just saw it, cool!

cftorres (approved):

PR #79, opened 2020-07-27
Display author name in NotePreview 

### LINKED ISSUE #21: Track prestige

keelywan:
- Moved `constructUserFromSqlResult` from `UserRegistrationServlet` to the User Service class and added `getUserById` function
- Added two getter methods in the User model 
  - `getDisplayName`
  - `getPoints`
## GetAuthorInfoServlet 
- Calls `getUserById` function to retrieve all the user's information
- Returns only the necessary data (display name and points) so the front end can't see private data like user_id

## NotePreview 
- Display's the author's name
preview1

- When the display name is hovered over, a tooltip will pop up showing the number of points that user has
preview2


Aradhya-B (approved):
LGTM :)

andres-holguin (approved):
Looks good!

src/main/java/com/google/starfish/servlets/GetAuthorInfoServlet.java
```diff
@@ -0,0 +1,49 @@
...
+        res.getWriter().println(json);
+      }
+    } catch (SQLException ex) {
+      res.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+      res.getWriter().println("INTERNAL SERVER ERROR: " + ex);
+    }
+  }
+
+  /** Converts an object to JSON */
+  private String convertObjectToJSON(HashMap<String, String> user) {
```
cftorres, line 45: I see this kind of logic in a lot of your servlets-- maybe you could use from Util file (I think Aradhya made this in a recent PR)?

cftorres (approved):
Nice-- very clean!

src/main/java/com/google/starfish/servlets/GetAuthorInfoServlet.java
Aradhya-B: Yep, I think we can make a new PR that includes a lot of these duplicated methods inside the `Utils` class once #72 gets merged!

nickKelman (approved):

PR #80, opened 2020-07-28
This adds unit tests for the LabelService, UserService, and NoteService classes

### LINKED ISSUE #81: Unit testing for Service Classes

Aradhya-B:
# Description
Using the [DBSetup unit testing framework](http://dbsetup.ninja-squad.com/), we can do unit testing on functions that need to speak to our SQL database. A new DB, `starfish_test` has been setup in the Cloud SQL console for this purpose.
Before every test is run, we clear the entire DB, insert a user, a school label, and a course label to serve as reference data (this is all done as `CommonOperations`) and then add any additional data needed for a given class's tests. Constants that are used throughout the various unit tests, as well as the DB DataSource, are stored in a `Constants` class. 
# Changes
- Remove type enum from Label model because it was unused and type safety is being checked already by the LabelService class Type enum
- Add getters for useful private fields in the models
- Remove the `getRowById` methods from the TableService class because they were returning a result set and were unusable due to data leakage
- Add methods to multiple service classes to get that service's model by an id
- Add unit tests to test the public methods of the NoteService, UserService, and LabelService class
# All Tests Passing
image


andres-holguin (approved):
Nice 👌

keelywan (approved):

cftorres (approved):

src/test/java/com/google/starfish/services/LabelServiceTest.java
```diff
@@ -0,0 +1,84 @@
...
+    dbSetup.launch();
+  }
+
+  /** Test that a school label can be inserted */
+  @Test
+  public void testInsertSchoolLabel() throws SQLException {
+    String testSchool = "mit";
+    labelService.insertSchoolLabel(pool, testSchool);
+    Label schoolLabel = labelService.getLabelByTitle(pool, testSchool);
+    assertTrue(schoolLabel.getTitle().equals(testSchool) && schoolLabel.getType().equals("School"));
```
cftorres: Actually, I would maybe separate these into two separate `assertTrue()` statements because if the test fails it'll be easier to know which one failed by separating them out (here and below)

Aradhya-B: Good point, I'll do that!

nickKelman (approved):
Really nice job testing this!

PR #83, opened 2020-07-28
Add vue routing to make SPA

### LINKED ISSUE #69: Landing Page

andres-holguin:
# Vue Routing
This PR adds vue routing to the application, making it a Single Page Application. This facilitates communication across its pages. In particular it enables the search bar in the landing page to produce a result on the search results page.

Starfish-Home-Router-V1

### DIFF
--- a/src/main/webapp/components/SearchBar.vue
+++ b/src/main/webapp/components/SearchBar.vue
@@ -71,8 +71,19 @@ module.exports = {
         .catch(error => {
           throw new Error(error);
         });
+    },
+  },
+  watch: {
+    // When $route changes, component is reused, so manually set new school and course and fetch results.
+    $route: function(to, from) {
+      this.school = to.query.school;
+      this.course = to.query.course;
+      this.fetchAndEmitResults();
     }
-  }
+  },
+  mounted: function() {
+    if (this.formIsValid) this.fetchAndEmitResults();
+  },
 }
 </script>
 
--- a/src/main/webapp/components/SearchBar.vue
+++ b/src/main/webapp/components/SearchBar.vue
@@ -60,10 +60,10 @@ module.exports = {
     },
     fetchAndEmitResults: async function() {
       let url = new URL("/search", window.location.href);
-      url.searchParams.set('school', school);
-      url.searchParams.set('course', course);
+      url.searchParams.set('school', this.school);
+      url.searchParams.set('course', this.course);
 
-      fetch(url)
+      return fetch(url)
         .then(response => response.json())
         .then(result => {
           this.$emit('searchresult', result);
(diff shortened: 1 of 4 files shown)


Aradhya-B (approved):
Looks great! Is `HomePage.vue` really the landing page at this point? Could the component made in #85 be put into `HomePage.vue`?

andres-holguin:
> Looks great! Is `HomePage.vue` really the landing page at this point? Could the component made in #85 be put into `HomePage.vue`?

That can definitely be done :) #85 can have its HTML converted to a template, and its CSS can be moved into that same component.

keelywan (approved):
This looks really good! I'll move the HTML in #85 into a Vue component 😄

src/main/webapp/components/SearchBar.vue
```diff
@@ -27,36 +27,63 @@
...
+      });
+    },
+    onInvalidForm: function () {
+      this.error = "Please enter at least one search criterion.";
+    },
+    fetchAndEmitResults: async function() {
       let url = new URL("/search", window.location.href);
-      url.searchParams.set('school', school);
-      url.searchParams.set('course', course);
+      this.school && url.searchParams.set('school', this.school);
```
cftorres, line 63: Perhaps I just don't know about this syntax, but why do you need to use this here when you're just calling a setter?

src/main/webapp/index.html
```diff
@@ -6,22 +6,46 @@
...
+
+  // 2. Define some routes
+  const routes = [
+    { path: '/', component: Home },
+    { path: '/search', component: Search }
+  ]
+
+  // 3. Create the router instance and pass the `routes` option
+  const router = new VueRouter({
+    routes // short for `routes: routes`
```
cftorres: nit: if you're likely not going to add anything to this object, especially since it's just a shorthand, I think you can include this inline above

cftorres (approved):
awesome!

src/main/webapp/components/SearchBar.vue
andres-holguin: I was having an issue where if the school was null, it would set the url search param for school=null, which the back end would read as school = "null" and then try to look for schools named null XD this makes sure null doesn't end up in the query string.

nickKelman (approved):
Awesome change.

src/main/webapp/index.html
andres-holguin: Makes sense, I can make that work :)

src/main/webapp/components/SearchBar.vue
cftorres: oh interesting!

PR #84, opened 2020-07-28
Move validateUser method to Utils class

keelywan:
## ValidationService
- `validateUser` method to check the HttpServletRequest's cookies and session
- Used by servlets that require the user to be logged in
  - FavoriteNoteServlet
  - HandleNotesServlet 
  - UploadGNoteServlet


Aradhya-B (changes requested):
The contents of the class looks good but I don't think this should be "service" class per-say. Design wise, service classes are meant to be the controllers (i.e. controllers in the MVC architecture) for the different models in the RDBMS (i.e. models in the MVC architecture) so I think this logic fits better inside a `Utils` in the `com.google.starfish.servlets` package where it is primarily being used.

Check out the `Utils` class I made in  #72. I added some util methods for the Search and TrendingNotes servlets. Since that isn't merged yet, I think you could make a Utils class from scratch and just add `validateUser` as a static method and then instead of making a new instance of `ValidationService` everywhere, just call directly using `Utils.validateUser()` since it'll be in the same package as the servlets. There shouldn't be any merge conflicts with what I have in #72 but even if there are, it should be easy to resolve.

Let me know if you have any questions!

Aradhya-B (approved):
LGTM, thanks for making the change I requested :)!

andres-holguin (approved):
LGTM

cftorres (approved):
Nice refactoring!

PR #86, opened 2020-07-29
This moves JSON converters from servlets to the Utils class

Aradhya-B:
Simple functions that convert a data structure to JSON are moved to the Utils class and called as static methods from servlets that need that functionality. The 4 methods, `convert2DArrayToJSON`, `convertArrayToJSON`, `convertHashMapToJSON`, and `convertObjectToJSON` can now also all be overloaded to handle different types.


andres-holguin (approved):
This looks good. I wonder why Gson's toJson() method isn't static 🤔

keelywan (approved):

cftorres (approved):
Nice!

nickKelman (approved):

PR #88, opened 2020-07-29
This adds unit tests for the FavoriteNoteService and MiscNoteLabelService classes

### LINKED ISSUE #81: Unit testing for Service Classes

Aradhya-B:
# Description
Unit tests are added for the `getMostUsedMiscLabels()` and `getMiscLabelsByNoteId()` methods of the `MiscNoteLabelService` class, for the `getFavoriteNotesByUserId()` and `getNumFavoritesByNoteId()` methods of the `FavoriteNoteService` class, and an additional 5 distinct unit tests for the `getTrendingNotesBySchoolOrCourse()` method of the `FavoriteNoteService` class. This makes for 9 additional unit tests, bringing the total up to 24.
 
The 5 distinct unit tests for `getTrendingNotesBySchoolOrCourse()` test various ways the method could be called, covering all 4 levels of `Recency` in the case where `school` and `course` are both `null`, and one final unit test to cover the case where `school` and `course` are not `null`.

The data inserted into the test database before the tests in addition to the reference data of 1 User, 1 School Label, and 1 Course Label was deemed to be the minimum required data entry to sufficiently test each class's methods. In particular, for the `FavoriteNoteServiceTest` class, either a high number of `Notes` or high number of `Users` could have been used to add multiple favorites on single notes to sufficiently test getting trending notes by `Recency` but a higher number of `Notes` was chosen.
# All Tests Passing
image


keelywan (approved):
Great testing!

andres-holguin (approved):

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
```diff
@@ -0,0 +1,300 @@
...
+  private String LABELS = constants.LABELS;
+  private String FAVORITE_NOTES = constants.FAVORITE_NOTES;
+  private String USERS = constants.USERS;
+  private DataSource pool = Constants.pool;
+  private FavoriteNoteService favoriteNoteService = new FavoriteNoteService();
+
+  private final Date today = favoriteNoteService.getDateBasedOnRecency(Recency.TODAY);
+  private final Date thisWeek = favoriteNoteService.getDateBasedOnRecency(Recency.THIS_WEEK);
+  private final Date thisMonth = favoriteNoteService.getDateBasedOnRecency(Recency.THIS_MONTH);
+  private final Date allTime = favoriteNoteService.getDateBasedOnRecency(Recency.ALL_TIME);
```
cftorres: I'm admittedly not as familiar anymore with JUnit testing, but...  Typically, access modifiers should not be changed just for tests, and it seems like you've changed the access modifier in the class above to be `protected` so that you can use the private method here.  Those access modifiers should be used for the sole purpose of meeting the needs of the API as designed to interact with other classes, etc.  

Regardless, when testing, data like this used in an assertion should actually likely be hardcoded rather than derived from some method.  You can imagine that if the method is faulty, testing some functionality that uses it would return the same faulty result as this expected value, and so, your test for this functionality would pass even if the method is faulty.  Rather, it would be much clearer, if it was made explicitly clear what values you're using to make assertions about expected vs. derived (from whatever functionality you're testing).

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
```diff
@@ -0,0 +1,300 @@
...
+  public void prepare() throws Exception {
+    if(!runTests) throw new Exception("Wrong Test Database Name");
+    Operation operation = getBeforeTestOperation();
+    DbSetup dbSetup = new DbSetup(new DataSourceDestination(pool), operation);
+    dbSetup.launch();
+  }
+
+  /** Test to see if we can get all notes favorited by a single user */
+  @Test
+  public void testGettingFavoriteNotesByUserId() throws SQLException {
```
cftorres: Ideally, this test would actually verify that this returns the entire array of notes as you would expect to receive, not just the length (that can perhaps just be a test to preface checking the entire value because if that's not true, then you needn't bother checking for the values).

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
```diff
@@ -0,0 +1,300 @@
...
+    Object[][] trendingNotesToday = favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, Recency.TODAY, null, null);
+    // We should be getting all 7 notes irrespective of whether they are trending or not
+    if (trendingNotesToday.length != 7) throw new Exception("Retrieved wrong number of notes: " + trendingNotesToday.length);
+    int numTrendingNotes = 2;
+    Note mostTrendingNote = (Note) trendingNotesToday[0][0];
+    long numFavoritesMostTrendingNote = (long) trendingNotesToday[0][1];
+    Note leastTrendingNote = (Note) trendingNotesToday[numTrendingNotes - 1][0];
+    long numFavoritesLeastTrendingNote = (long) trendingNotesToday[numTrendingNotes - 1][1];
+    long numFavoritesFirstNoteWithNoFavorite = (long) trendingNotesToday[numTrendingNotes][1];
+    assertTrue(mostTrendingNote.getId() == 2);
```
nickKelman: It's good practice to have unit tests be atomic (Test 1 thing each). See if you can utilize a beforeEach statement and break this out into separate tests.

nickKelman: Applies to the tests below as well.

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
```diff
@@ -0,0 +1,300 @@
...
+    assertTrue(numFavoritesMostTrendingNote == 2);
+    assertTrue(leastTrendingNote.getId() == 1);
+    assertTrue(numFavoritesLeastTrendingNote == 1);
+    assertTrue(numFavoritesFirstNoteWithNoFavorite == 0);
+  }
+
+  /** Test to see if we can correctly retrieve the most trending notes this week */
+  @Test 
+  public void testGettingTrendingNotesThisWeek() throws SQLException, Exception {
+    Object[][] trendingNotesThisWeek = favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, Recency.THIS_WEEK, null, null);
```
nickKelman: I see a lot of similar logic at the start of each test. Could this logic be consolidated into a helper function?

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
```diff
@@ -0,0 +1,300 @@
...
+    long numFavoritesOfNoteTwo = favoriteNoteService.getNumFavoritesByNoteId(pool, 2);
+    assertTrue(numFavoritesOfNoteOne == 1);
+    assertTrue(numFavoritesOfNoteTwo == 2);
+  }
+
+  /** Test to see if we can correctly retrieve the most trending notes today */
+  @Test
+  public void testGettingTrendingNotesToday() throws SQLException, Exception {
+    Object[][] trendingNotesToday = favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, Recency.TODAY, null, null);
+    // We should be getting all 7 notes irrespective of whether they are trending or not
```
cftorres: (Here and below) I'm a bit confused by this statement that you should be getting *all* notes irrespective of whether they're trending or not-- perhaps the name of the method then is misleading because you say you're getting trending notes, specifically?  Also, aren't you necessarily limiting the number of notes you're getting by having a date limit, so why would you be getting all?

nickKelman (approved):

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
```diff
@@ -0,0 +1,300 @@
...
+    if (trendingNotesToday.length != 7) throw new Exception("Retrieved wrong number of notes: " + trendingNotesToday.length);
+    int numTrendingNotes = 2;
```
cftorres: (Here and below) I'm also not sure where this random number is coming from.  I suppose you just know off the top of your head that this number should be two ?  It might be helpful if you created your fake data such that you had an array like `TRENDING_NOTES_TODAY = [note, note]`, then `TRENDING_NOTES_THIS_WEEK = [TRENDING_NOTES_TODAY] + [note, note]` or so so that here you can just say `numTrendingNotes = TRENDING_NOTES_TODAY.length` and it would be much clearer

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
```diff
@@ -0,0 +1,300 @@
...
+    Note mostTrendingNote = (Note) trendingNotesToday[0][0];
+    long numFavoritesMostTrendingNote = (long) trendingNotesToday[0][1];
+    Note leastTrendingNote = (Note) trendingNotesToday[numTrendingNotes - 1][0];
+    long numFavoritesLeastTrendingNote = (long) trendingNotesToday[numTrendingNotes - 1][1];
+    long numFavoritesFirstNoteWithNoFavorite = (long) trendingNotesToday[numTrendingNotes][1];
+    assertTrue(mostTrendingNote.getId() == 2);
+    assertTrue(numFavoritesMostTrendingNote == 2);
+    assertTrue(leastTrendingNote.getId() == 1);
+    assertTrue(numFavoritesLeastTrendingNote == 1);
+    assertTrue(numFavoritesFirstNoteWithNoFavorite == 0);
```
cftorres: similar to a comment above, I think a better test would be to actually just check the array, that the note values correspond to every note you would expect in the array.

src/test/java/com/google/starfish/services/MiscNoteLabelServiceTest.java
```diff
@@ -0,0 +1,134 @@
...
+  private Constants constants = new Constants();
+  private String LABELS = constants.LABELS;
+  private String NOTES = constants.NOTES;
+  private String MISC_LABELS = constants.MISC_LABELS;
+  private DataSource pool = Constants.pool;
+  private MiscNoteLabelService miscNoteLabelService = new MiscNoteLabelService();
+
+  private final String miscLabelOne = "best";
+  private final String miscLabelTwo = "hard";
+  private final String miscLabelThree = "math";
```
cftorres: (Here and elsewhere) when you have hard-coded values like this, I think it helps readability to make these styled as constants (i.e. all caps) so that we can clearly distinguish in the test when you have a derived value (the thing you're usually testing) vs. an expected value

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
Aradhya-B: That makes a lot of sense, I suppose I assumed that a unit test only gets as granular as "test one function" but I guess it's really about "testing one aspect of a function". Like any one function should only have a singular functionality but there might be different aspects of that functionality (like in this case) that should be tested independently.

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
Aradhya-B: Definitely, I think a lot of it can be extracted into a helper function that sets instance variables of the entire test class and that's called in the Before decorated prepare statement.

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
Aradhya-B: This makes so much sense and I definitely thought about this when I changed the access modifier of the `getDateBasedOnRecency` method to work in the test class. I think I was trying to make the date generated dynamic, i.e. `today` would always be in reference to the actual date today when the test is run, but in hindsight, there really isn't any need for that and the way I've done it here, if the `getDateBasedOnRecency` method was faulty, I would've never known just from these unit tests because they implicitly assume that it's 100% functional. I'll hardcode the dates!

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
Aradhya-B: Ahh this makes sense, the test does definitely feel incomplete without checking the actual values of the notes, even if the test passes, there's a lingering feeling of "what if the notes retrieved were wrong?" I'll make a change for this!

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
Aradhya-B: I believe we introduced this functionality to the `getTrendingNotesBySchoolOrCourse` method to work as we expected with the search servlet. Basically, irrespective of the timespan somebody is seeking "trending" notes, we want the backend to return all the note data sorted by the number of favorites in that timespan, and any note that isn't "trending" would have a number of favorites in that timespan of 0, but would still be returned because it would still technically fit the user's search query. Like, this method isn't being used in the context of a user searching specifically for trending notes in a given timespan, but a user just searching for notes in general and this method allows us to internally rank those notes. The notes that aren't "trending" just move to the bottom of the array and then it's really easy for the frontend to filter those notes out by removing notes with a 0 count on num favorites in the timespan if that's the functionality desired there.

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
Aradhya-B: This number basically represents how many notes have a row in the favorite_notes table in the given timespan, so like in the `getBeforeTestOperation` method, our operation would make sure 2 notes are favorited "today". It also is making sure that 2 additional notes are favorited in the timespan of "this week", but since "this week" includes "today", the total number of trending notes "this week" is 2 + 2 = 4. Then same thing for "this month" which gets us 6. And then for "all time" there's one additional note favorited beyond the scope of "this month" so the number of trending notes all time becomes 7.

I'm not sure having arrays for the dummy data would be the best move here in terms of efficiency because that would mean we'd have to construct all those notes separately and then retrieve all those fields when we're setting values to insert in the SQL tables, all when we know from the beginning what data needs to be inserted into the tables.

Aradhya-B: What I think I could do is make all these variables constants of the test class and leave a comment explaining that these numbers mean how many notes in total are favorited in the given timespan, so for anybody in the future looking at these tests, they'd know that the correlation of these constants is to the number of favorites being inserted as dummy data into the favorite_notes table in the `getBeforeTestOperation` method.

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
Aradhya-B: So basically are you saying that instead of having 5 of these separate assertTrue statements, I could basically just have one that checks if some helper method returns true where the helper method checks the note values of the passed array for the given test?

src/test/java/com/google/starfish/services/MiscNoteLabelServiceTest.java
Aradhya-B: Definitely agree with this, my mistake for not making the constants all caps from the beginning!

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
```diff
@@ -90,9 +93,7 @@ public void testGettingFavoriteNotesByUserId() throws SQLException, Exception {
       throw new Exception("Expected 3 notes for user " + SECOND_NEW_USER_ID + ", got: " + favoriteNotes.length);
     } 
     Arrays.sort(favoriteNotes, COMPARE_NOTES_BY_ID);
-    assertTrue(favoriteNotes[0].getId() == 2);
-    assertTrue(favoriteNotes[1].getId() == 4);
-    assertTrue(favoriteNotes[2].getId() == 6);
```
cftorres: Maybe add a comment (here and below) making clear that `favoriteNotes` is set in `setAllTrendingNotes()` call when `prepare()` gets executed because I was wondering where you were deriving this value from.  Or, honestly, I wouldn't even set them there, I would set each of them in the test that requires them since they're not needed anywhere else and it makes it a little more explicit exactly what's happening.

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
```diff
@@ -0,0 +1,423 @@
...
+    notesOrderedByTrendingThisWeek = extractNotesFrom2DArray(
+        favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, Recency.THIS_WEEK, null, null));
+    notesOrderedByTrendingThisMonth = extractNotesFrom2DArray(
+        favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, Recency.THIS_MONTH, null, null));
+    notesOrderedByTrendingAllTime = extractNotesFrom2DArray(
+        favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, Recency.ALL_TIME, null, null));
+    notesOrderedByTrendingAllTimeFiltered = extractNotesFrom2DArray(
+        favoriteNoteService.getTrendingNotesBySchoolOrCourse(pool, Recency.ALL_TIME, NEW_SCHOOL, NEW_COURSE));
+
+    if (notesOrderedByTrendingToday.length != 7) {
```
cftorres: (here and below) can't you use NOTES_EXPECTED_TODAY.length instead of having a magic number here?

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
```diff
@@ -0,0 +1,423 @@
...
+  private Date TODAY = getToday();
+  private Date THIS_WEEK = getThisWeek();
+  private Date THIS_MONTH = getThisMonth();
+  private Date ALL_TIME = getAllTime();
+
+  // These variables represent the total number of rows that should be inserted 
+  // into the FAVORITE_NOTES table where the data is as indicated.
+  private final int NUM_TRENDING_NOTES_TODAY = 2;
+  private final int NUM_TRENDING_NOTES_THIS_WEEK = NUM_TRENDING_NOTES_TODAY + 2;
+  private final int NUM_TRENDING_NOTES_THIS_MONTH = NUM_TRENDING_NOTES_THIS_WEEK + 2;
```
cftorres: It doesn't look like you use these anymore?

cftorres (approved):
Awesome work on this CL-- looks a lot cleaner!

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
Aradhya-B: Oh yep you're right! Will make that change

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
Aradhya-B: Yeah I was thinking about this too after I'd added the `setAllTrendingNotes()` function. Originally it was added to abstract some code out of the more complex and verbose test, but now that the tests are all very simplified, it makes more sense just to set them in the test themselves. I'll make that change!

src/test/java/com/google/starfish/services/FavoriteNoteServiceTest.java
Aradhya-B: Yep you're right, good catch! I'll get rid of them

PR #85, opened 2020-07-28
Create landing page

keelywan:
## Landing Page
Screenshot 2020-07-29 at 3 56 13 PM

Screenshot 2020-07-30 at 12 36 53 PM

Screenshot 2020-07-30 at 12 37 13 PM

## Footer
Screenshot 2020-07-30 at 12 37 43 PM

## Search Bar 
- Modified the search bar so it takes two props: `isDark` and `showSearchButton`
- This allows for the different appearance of the search bar on the landing page (passes `isDark="true"` and `showSearchButton="false"` to the component)

PR #89, opened 2020-07-29
Add trending notes category

### LINKED ISSUE #68: Get Trending Notes

andres-holguin:
###### Best reviewed after #83 has closed or filtered by commits after and including 067ee4d. Try [this link](https://github.com/googleinterns/step102-2020/pull/89/files/21d4b5c1c6e5c87893690ce90529427ca0d280da..cacd0c1fdc29bb6e7ea36c90db5b663b71b45e6e) or a variation instead of clicking on Files changed.
This PR implements the a trending notes category based on the back-end work done by @Aradhya-B  in #72 
image
Since this new backend introduced several bugs in the frontend, this PR also involves
- Fixing how noteData from the back end is parsed by the front end
- Fixing how the dateCreated string from the back end is parsed by the front end for sorting, and then stringified for display.


src/main/webapp/components/NoteGridCollection.vue
```diff
@@ -50,6 +50,7 @@ module.exports = {
       this.nextTick(() => this.selected = null)
       // Filter by date, matching terms, and sort with compareFunc
       return this.noteData
+        .map(this.parseData)
```
Aradhya-B, line 53: Since every set of `noteData` passed to the `NoteGridCollection` isn't a nested array of notes/numFavoritesInTimespan (like on the profile page where we want to just pass a 1D array of notes to the `NoteGridCollection`), maybe we can introduce a prop to `NoteGridCollection`, something like "trendingData" or something which is a boolean that indicates whether or not this `noteData` is a nested array or 1D array. Then only if `trendingData` is true, we'll map the data using the `parseData` function.

src/main/webapp/views/HomePage.vue
```diff
@@ -0,0 +1,121 @@
...
+  .main-nav {
+    font-size: 0.5em;
+    padding: 0;
+  }
+  #laptop {
+    width: 400px;
+    height: auto;
+  }
+}
+</style>
```
Aradhya-B: Could we possibly make this file blank or even delete it in this PR so the Landing/HomePage made in #85 can get merged with no conflicts?

Aradhya-B (changes requested):
This looks great! Awesome job refactoring the frontend given the changes to the API. Requesting changes primarily for the NoteGridCollection to ensure that it can take either a 2D or 1D array of notes and handle accordingly (see comment).

src/main/webapp/views/HomePage.vue
andres-holguin: Sure, shouldn't be a problem. This is actually a change from the vue router PR #83, so I'd have to either remove it from both PRs or just resolve the merge conflicts when they occur.

keelywan (approved):
Looks good, same thing that Aradhya mentioned above about getting rid of HomePage.vue (possibly SearchBar.vue too because I think I saw that in another PR?)

andres-holguin:
> Looks good, same thing that Aradhya mentioned above about getting rid of HomePage.vue (possibly SearchBar.vue too because I think I saw that in another PR?)

Since this PR depends on #83, when reviewing file changes it's best to view after commit 067ee4d since you'd be seeing all of #83's changes otherwise. Try [this link](https://github.com/googleinterns/step102-2020/pull/89/files/21d4b5c1c6e5c87893690ce90529427ca0d280da..cacd0c1fdc29bb6e7ea36c90db5b663b71b45e6e) instead of clicking on Files Changed.

src/main/webapp/components/NoteGridCollection.vue
andres-holguin: Okay, just pushed a commit that seems to implement what you had an mind and still works.

Aradhya-B: Thanks for making the change, it looks great and I think what you've done is better than the suggestion I made to add a prop because this way the user of the component doesn't need to think about what kind of data they're passing, they can just pass one of the expected formats and it just works!

Aradhya-B (approved):
LGTM, thanks for making that non-trivial change :)!

src/main/webapp/views/HomePage.vue
```diff
@@ -0,0 +1,121 @@
...
+}
+
+/* COLORS */
+.navy-blue {
+  background-color: #004aad;
+}
+
+/* NAV */
+.main-nav {
+  display: flex;
```
nickKelman: Keely was keeping these properties sorted, I'd sync up with her to see if its what you all want to continue doing.

nickKelman:
This looks great, really nice work.

nickKelman (approved):

nickKelman (approved):

PR #90, opened 2020-07-30
This adds a profile page where users can view their points, favorited notes, and uploaded notes

### LINKED ISSUE #13: View Profile page

Aradhya-B:
# If Signed In
image
# If Not Signed In
image

TODO: Route between pages using Vue Router introduced by @andres-holguin  in #83


src/main/webapp/views/ProfilePage.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <v-app>
+    <navbar @setuser="onSetUser" @signedin="onSignIn"></navbar>
+    <div v-if="!signedIn">
+      <h1>Please sign in</h1>
+    </div> 
+    <div v-else> 
+        <profile-header :points="user.points"
```
keelywan: nit: indentation (I think it's indented twice) and same thing for below

andres-holguin (approved):
Very Nice :)

keelywan (approved):
Looks good!

src/main/webapp/views/ProfilePage.vue
Aradhya-B: Ah you're right, good catch!

nickKelman (approved):

cftorres (approved):

PR #92, opened 2020-08-04
Integration and bug fixes 

keelywan:
## Small Fixes 
- Fix `source_url` typo in FavoriteNoteService
- Computed `isGnote` value
- Add a null check for `miscLabels` so the `"null"` string won't be added to the database

## Additions/Modifications
- Add `doGet` to DownloadNoteServlet to return the number of downloads a note has
- Modify `doGet` in FavoriteNoteServlet to return the number of favorites a note has, along with whether or not the note is favorited by the user
- Modify sign in on the navbar and add the ProfilePage component to the navbar dropdown


Aradhya-B (approved):
WOOO, finally integrated!!

andres-holguin (approved):
😁

nickKelman (approved):

